PR #2, opened 2024-07-30
update secret names

prabash-moj:
update secret names to fix start up issue


helm_deploy/values-dev.yaml
```diff
@@ -14,5 +14,17 @@ generic-service:
...
+  #   [name of kubernetes secret]:
+  #     [name of environment variable as seen by app]: [key of kubernetes secret to load]
+
+  namespace_secrets:
+    hmpps-contacts-ui:
+      SIGN_IN_CLIENT_ID: "SIGN_IN_CLIENT_ID"
+      SIGN_IN_CLIENT_SECRET: "SIGN_IN_CLIENT_SECRET"
+      SYSTEM_CLIENT_ID: "SYSTEM_CLIENT_ID"
+      SYSTEM_CLIENT_SECRET : "SYSTEM_CLIENT_SECRET"
+
```
timharrison-moj, line 28: Sorry - I might have misled you here - these are not needed per env, as the base properties in values.yaml apply to all environments, and it reads the values from secrets in each env. So not needed in this file.

server/config.ts
```diff
@@ -76,8 +76,8 @@ export default {
         deadline: Number(get('HMPPS_AUTH_TIMEOUT_DEADLINE', 10000)),
       },
       agent: new AgentConfig(Number(get('HMPPS_AUTH_TIMEOUT_RESPONSE', 10000))),
-      apiClientId: get('API_CLIENT_ID', 'clientid', requiredInProduction),
-      apiClientSecret: get('API_CLIENT_SECRET', 'clientsecret', requiredInProduction),
+      apiClientId: get('SIGN_IN_CLIENT_ID', 'clientid', requiredInProduction),
+      apiClientSecret: get('SIGN_IN_CLIENT_SECRET', 'clientsecret', requiredInProduction),
       systemClientId: get('SYSTEM_CLIENT_ID', 'clientid', requiredInProduction),
```
timharrison-moj, line 81: We probably want the config variable names to match up too? 
`apiClientId` -> `signInClientId`
`apiClientSecret` -> `signInClientSecret` 
So that usage is consistent?

prabash-moj: added a fix for this now

timharrison-moj (approved):
LGTM

PR #3, opened 2024-07-30
fix dns host name issue

prabash-moj:
fix dns host name issue

### DIFF
--- a/helm_deploy/values-dev.yaml
+++ b/helm_deploy/values-dev.yaml
@@ -5,10 +5,10 @@ generic-service:
   replicaCount: 2
 
   ingress:
-    host: contacts-ui-dev.hmpps.service.justice.gov.uk
+    host: contacts-dev.hmpps.service.justice.gov.uk
 
   env:
-    INGRESS_URL: "https://contacts-ui-dev.hmpps.service.justice.gov.uk"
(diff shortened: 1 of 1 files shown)


helm_deploy/values-dev.yaml
```diff
@@ -5,7 +5,7 @@ generic-service:
   replicaCount: 2
 
   ingress:
-    host: contacts-ui-dev.hmpps.service.justice.gov.uk
+    host: contacts-dev.hmpps.service.justice.gov.uk
```
timharrison-moj, line 8: Could you make same change as this to the -preprod and -prod values too please? We dont actually need right now, but it will catch us out later if we don't. Thanks.

prabash-moj: I will make a note of that, im sure i will forget otherwise :)

helm_deploy/values-dev.yaml
```diff
@@ -5,7 +5,7 @@ generic-service:
...
 
   env:
     INGRESS_URL: "https://contacts-ui-dev.hmpps.service.justice.gov.uk"
```
timharrison-moj, line 11: Same change here too please... INGRESS_URL should be contacts-dev.xxx.xx.xx.

timharrison-moj (approved):
LGTM. I'll update the -preprod -prod files shortly

PR #5, opened 2024-07-31
PRC-4 Landing Page

rosariodiaferia-moj:
**Ticket**: https://dsdmoj.atlassian.net/browse/PRC-4

**Description**: 
Options on the tiled home page added using card component:
- Manage contacts
- Manage contact restrictions


logger.ts
```diff
@@ -4,6 +4,6 @@ import config from './server/config'
 
 const formatOut = bunyanFormat({ outputMode: 'short', color: !config.production })
 
-const logger = bunyan.createLogger({ name: 'Hmpps Contacts Ui', stream: formatOut, level: 'debug' })
+const logger = bunyan.createLogger({ name: 'Digital Prison Services', stream: formatOut, level: 'debug' })
```
prabash-moj, line 7: I think this should be still 'hmpps contacts UI' ?

package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "hmpps-contacts-ui",
   "version": "0.0.1",
-  "description": "Hmpps Contacts Ui",
+  "description": "Digital Prison Services",
```
prabash-moj, line 4: dont change the description as it should be unique to this service?

server/utils/nunjucksSetup.ts
```diff
@@ -11,7 +11,7 @@ export default function nunjucksSetup(app: express.Express): void {
   app.set('view engine', 'njk')
 
   app.locals.asset_path = '/assets/'
-  app.locals.applicationName = 'Hmpps Contacts Ui'
+  app.locals.applicationName = 'Digital Prison Services'
```
prabash-moj, line 14: not sure if we should change the application name here?

logger.ts
prabash-moj: I think this should still stay as it was before because it is the name for the logger for this module, we will see this see logs on the console with this name hence easily stand out from the other apps

package.json
rosariodiaferia-moj: Hmpps Contacts Ui added back

server/utils/nunjucksSetup.ts
rosariodiaferia-moj: Hmpps Contacts Ui added back

prabash-moj (approved):

PR #12, opened 2024-08-12
PRC-34 Create Search For A Prisoner Form Field And Validation

rosariodiaferia-moj:
**Ticket**: https://dsdmoj.atlassian.net/browse/PRC-34
**Description**: Search prisoner form and validation

**Video**:

https://github.com/user-attachments/assets/b57e3236-505f-43e3-a4d9-8469049e089b

### DIFF
--- a/integration_tests/e2e/signIn.cy.ts
+++ b/integration_tests/e2e/signIn.cy.ts
@@ -1,4 +1,4 @@
-import IndexPage from '../pages/index'
+import ContactsPage from '../pages/contacts'
 import AuthSignInPage from '../pages/authSignIn'
 import Page from '../pages/page'
 import AuthManageDetailsPage from '../pages/authManageDetails'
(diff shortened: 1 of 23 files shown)


integration_tests/e2e/signIn.cy.ts
```diff
@@ -21,36 +21,36 @@ context('Sign In', () => {
 
   it('User name visible in header', () => {
     cy.signIn()
-    const indexPage = Page.verifyOnPage(IndexPage)
-    indexPage.headerUserName().should('contain.text', 'J. Smith')
+    const contactsPage = Page.verifyOnPage(ContactsPage)
+    contactsPage.headerUserName().should('contain.text', 'J. Smith')
   })
 
   it('User name visible in header', () => {
```
prabash-moj, line 28: can this test be removed as it is same as the previous test ?

prabash-moj (dismissed):

integration_tests/e2e/signIn.cy.ts
rosariodiaferia-moj: Yes, I duplicated by mistake. It's been removed now

server/routes/urls.ts
```diff
@@ -0,0 +1,4 @@
+const BASE_SEARCH_URL = '/search'
+const SEARCH_PRISONER_URL = `${BASE_SEARCH_URL}/prisoner`
+
```
timharrison-moj, line 3: There is a `config.ts` module in most services to get the environment variable values, or default the value for test running. I think it would be best to stick with the convention for providing these env values?

prabash-moj (approved):

PR #16, opened 2024-08-23
PRC-54 : create a new contact

prabash-moj:
[RPC-54](https://dsdmoj.atlassian.net/browse/PRC-60) : want to capture and store new contact details, including Title, First Name, Middle Name, and Surname,

Test evidence: 
image


server/utils/utils.ts
```diff
@@ -21,3 +24,46 @@ export const initialiseName = (fullName?: string): string | null => {
   const array = fullName.split(' ')
   return `${array[0][0]}. ${array.reverse()[0]}`
 }
+
+export const parseDate = (date: string, fromFormat = 'yyyy-MM-dd') => {
```
james-baker-moj, line 28: We should add some tests for these functions.

server/routes/contacts/create/createContactController.ts
```diff
@@ -0,0 +1,11 @@
+import { Request, Response } from 'express'
+import { Page } from '../../../services/auditService'
+import { PageHandler } from '../../../interfaces/pageHandler'
+
+export default class CreateContactController implements PageHandler {
+  public PAGE_NAME = Page.SEARCH_PRISONER_CONTACT_NAME_PAGE
+
+  GET = async (req: Request, res: Response): Promise<void> => {
+    res.render('contacts/viewCreateContact')
```
james-baker-moj, line 9: Maybe rename the view to just `createContact`? I find the name `viewCreateContact` a bit confusing.

server/routes/contacts/create/viewCreateContact.njk
```diff
@@ -0,0 +1,92 @@
...
+                    value: formResponses.firstName or session.journey.bookAVideoLink.prisoner.firstName
+                }) }}
+
+                {{ govukInput({
+                    id: "lastName",
+                    name: "lastName",
+                    label: { text: "Last name" },
+                    classes: 'govuk-!-width-one-third',
+                    errorMessage: validationErrors | findError('lastName'),
+                    value: formResponses.lastName or session.journey.bookAVideoLink.prisoner.lastName
```
james-baker-moj, line 48: C&P from bookAVideoLink? I think we can remove reference to session for now.

server/utils/utils.ts
```diff
@@ -21,3 +24,46 @@ export const initialiseName = (fullName?: string): string | null => {
...
+  const separator = dateMatches[2]
+  const year = dateMatches[4]
+
+  const date = parse(datePickerDate, `dd${separator}MM${separator}${'y'.repeat(year.length)}`, startOfToday())
+  if (!isValid(date)) return new Date(NaN)
+
+  return date
+}
+
+export const simpleDateToDate = (date: { day: string; month: string; year: string }): Date =>
```
james-baker-moj, line 58: Some (most?) of these seem to be unused. I guess copied from BVLS? I would prefer we only import stuff we're actually using but if you need them imminently, it's fine to leave them as long we import the tests as well.

james-baker-moj (commented):
We should also put in the structure for cypress tests as a starting point here unless there is a separate ticket?

james-baker-moj:
Rebased with main and resolved conflicts

james-baker-moj (approved):
LGTM

PR #21, opened 2024-08-30
PRC-80: Add helm deploy required env vars and minor tidying updates.

timharrison-moj:
This will resolve the deployment issue.


james-baker-moj (approved):
LGTM

PR #23, opened 2024-08-30
PRC-56 : Session cleanup for create contact journey

james-baker-moj:
Prevent too many journeys from entering the session. If there are more than 5, remove the journey that was interacted with the longest ago.


timharrison-moj (approved):
LGTM

PR #24, opened 2024-09-02
PRC-56 : Call the create API

james-baker-moj:
Call the create API after setting the DOB. Initial import of contacts API types.


timharrison-moj (approved):
LGTM.

PR #31, opened 2024-09-04
PRC-82: Adding a resusable macro and config for a mini-profile for forms with a prisoner in context.

timharrison-moj:
This PR:

* Adds a prisonerImageService and prisonAPI client to get the prisoner image data to show.
* Adds a route so that the mini profile can access the image data (calling itself locally)
* Adds config for PRISON_API_URL and sets up the link to get to the prisoner profile page via DPS_HOME_PAGE_URL.
* Adds a macro, template and SCSS for the mini-profile, as a component.
* Trial run in the listContacts.njk form 
Screenshot 2024-09-04 at 11 41 01

Tests are being written.


rosariodiaferia-moj (approved):

server/views/partials/miniProfile/template.njk
```diff
@@ -0,0 +1,29 @@
...
+                     <span class="govuk-heading-s govuk-!-margin-bottom-0 govuk-!-padding-0">Date of birth</span>
+                     <span class="govuk-body" data-qa="mini-profile-dob">{{ person.dateOfBirth | formatDate }}</span>
+                </li>
+                <li>
+                    <span class="govuk-heading-s govuk-!-margin-bottom-0 govuk-!-padding-0">Establishment</span>
+                    <span class="govuk-body" data-qa="mini-profile-prison-name">{{ person.prisonName }}</span>
+                </li>
+                <li>
+                    <span class="govuk-heading-s govuk-!-margin-bottom-0 govuk-!-padding-0">Cell number</span>
+                    <span class="govuk-body" data-qa="mini-profile-cell-location">{{ person.cellLocation }}</span>
```
james-baker-moj, line 24: I have a feeling this is optional. Might want to check how it's presented, I've seen a few options across different services.

timharrison-moj: I've added an image in the comments showing how its presented. The designs we have are a bit inconsistent - i.e. date formats vary across the same journey, and mini-profile hardly has any info. I'm used some artistic licence to guess at what might be useful, and made things consistent.

server/views/pages/contacts/manage/listContacts.njk
```diff
@@ -28,6 +29,9 @@
     </div>
 </div>
 
+{# Test the mini profile macro #}
+{{ miniProfile(prisonerDetails, '/prisoner-image/' + prisonerDetails.prisonerNumber, digitalPrisonServicesUrl + '/prisoner/' + prisonerDetails.prisonerNumber) }}
```
prabash-moj, line 33: would be nice if the number of parameters could be reduced to one or two parameters may be personDetails would be enough, 

Could the url be moved into properties files and read from there, ''/prisoner-image/' and '/prisoner/'

PR #34, opened 2024-09-05
PRC-57: Support check answers update name flow

james-baker-moj:
Introduced flag to set once we are checking answers.  This is then used to redirect back to check answers straight from the enter name page. Also made it so the page is properly populated with either session data or form data if there was a validation error.

### DIFF
--- a/server/routes/contacts/create/enter-name/createContactEnterNameController.ts
+++ b/server/routes/contacts/create/enter-name/createContactEnterNameController.ts
@@ -6,10 +6,37 @@ import { CreateContactEnterNameSchemaType } from './createContactEnterNameSchema
 export default class EnterNameController implements PageHandler {
   public PAGE_NAME = Page.CREATE_CONTACT_NAME_PAGE
 
+  private TITLES = [
+    { value: '', text: '' },
+    { value: 'DAME', text: 'Dame' },
+    { value: 'FR', text: 'Father' },
+    { value: 'LORD', text: 'Lord' },
+    { value: 'MS', text: 'Ms' },
+    { value: 'RABBI', text: 'Rabbi' },
+    { value: 'REV', text: 'Reverend' },
+    { value: 'SIR', text: 'Sir' },
+    { value: 'SR', text: 'Sister' },
+    { value: 'MR', text: 'Mr' },
+    { value: 'BR', text: 'Brother' },
+    { value: 'DR', text: 'Dr' },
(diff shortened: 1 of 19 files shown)


server/routes/contacts/create/enter-name/createContactEnterNameController.ts
```diff
@@ -6,10 +6,37 @@ import { CreateContactEnterNameSchemaType } from './createContactEnterNameSchema
 export default class EnterNameController implements PageHandler {
   public PAGE_NAME = Page.CREATE_CONTACT_NAME_PAGE
 
+  private TITLES = [
```
james-baker-moj, line 9: I've moved this here temporarily as it's very clunky to set the selected value within the nunjucks. To be replaced by reference data soon hopefully

server/routes/contacts/create/enter-name/createContactEnterNameController.ts
```diff
@@ -6,10 +6,37 @@ import { CreateContactEnterNameSchemaType } from './createContactEnterNameSchema
...
     const { journeyId } = req.params
     const journey = req.session.createContactJourneys[journeyId]
-    res.render('pages/contacts/create/enterName', { journey })
+
+    const viewModel = {
+      journey,
+      titleOptions: this.getSelectedTitleOptions(res.locals?.formResponses?.title ?? journey?.names?.title),
+      lastName: res.locals?.formResponses?.lastName ?? journey?.names?.lastName,
+      firstName: res.locals?.formResponses?.firstName ?? journey?.names?.firstName,
+      middleName: res.locals?.formResponses?.middleName ?? journey?.names?.middleName,
```
timharrison-moj, line 37: Wondering about this part? Nunjucks templates can directly access both the journey and formResponses, so can do this logic on setting values   e.g.   `value = res.locals?.formResponses?.lastName ?? journey?.names?.lastName`, though it is fairly neat to just provide a `viewModel` for each form.  Trade off between a) Building a view model in handlers with extra code, or b) Reducing code in handlers, but increasing it in nunjucks. Ultimately, a choice..

james-baker-moj: Thanks, I was also unsure whether to do it here or in the template itself. I had 1 eye on DOB where we have to convert back from a date into the day, month, year fields and that felt like it'd be much easier in the controller.
Also search is currently using this pattern.

PR #35, opened 2024-09-05
 PRC-40 Prisoner Contact List

rosariodiaferia-moj:
**Ticket**: https://dsdmoj.atlassian.net/browse/PRC-40
**Description**: 
- Tab component
- Contacts list table
- Unit Test
- Integration Test

**Video:**

https://github.com/user-attachments/assets/fb41ce66-5018-4793-a50d-d601981a76a4


server/routes/contacts/manage/list/listContactsController.test.ts
```diff
@@ -56,15 +58,87 @@ afterEach(() => {
...
+    prisonerSearchService.getByPrisonerNumber.mockResolvedValue(TestData.prisoner())
+    contactsService.getPrisonerContacts.mockReturnValue(contactsList)
 
     const response = await request(app).get(`/contacts/manage/list/${journeyId}`)
 
+    // Prisoner
     expect(response.status).toEqual(200)
+    expect(response.text).toContain('John')
+    expect(response.text).toContain('Smith')
+    expect(response.text).toContain('2 April 1975')
```
timharrison-moj, line 97: We could be a bit more specific about which fields/columns contain which values?

james-baker-moj: I just added `cheerio` which makes this a bit easier as you can use jquery type selectors.

```
    const $ = cheerio.load(response.text)
    expect($('#firstName').val()).toStrictEqual('first')
    expect($('#middleName').val()).toStrictEqual('middle')
    expect($('#lastName').val()).toStrictEqual('last')
    expect($('#title').val()).toStrictEqual('MR')
```

server/utils/nunjucksSetup.ts
```diff
@@ -58,4 +58,41 @@ export default function nunjucksSetup(app: express.Express): void {
   njkEnv.addFilter('formatDate', formatDate)
   njkEnv.addGlobal('DPS_HOME_PAGE_URL', config.serviceUrls.digitalPrison)
   njkEnv.addFilter('pluralise', (word, count, plural = `${word}s`) => (count === 1 ? word : plural))
+
+  njkEnv.addFilter('createContactsListRows', contactList => {
+    const activeContactsRows: Array<unknown> = []
+    contactList.forEach((item: Record<string, Date>) => {
+      let addressLine = ''
+      addressLine += item.flat ? `${item.flat}<br />` : ''
```
timharrison-moj, line 66: Would prefer we didn't embed HTML <br/> tags into code, or inline functions like this as nunjucks filters because its hard to test them.  It also mixes code with the presentation layer, which is a generally bad idea. I think there are better ways of doing this using SCSS and an address macro that we can re-use elsewhere? I can help with this.

timharrison-moj (commented):
Couple of comments added.

server/routes/contacts/manage/list/listContactsController.test.ts
rosariodiaferia-moj: changes added using cheerio

server/utils/nunjucksSetup.ts
rosariodiaferia-moj: Code refactored using addressToLines from hmpps prisoner profile

server/utils/addressToLines.ts
```diff
@@ -0,0 +1,22 @@
...
+  city,
+  county,
+  postalCode,
+  country,
+}: Partial<PrisonerContactSummary>): string => {
+  let lineOne = [premise, street].filter(s => s).join(', ')
+  if (flat) {
+    lineOne = `Flat ${flat}, ${lineOne}`
+  }
+  const addressArray = [lineOne, area, city, county, postalCode, country].filter(s => s).join('<br />')
```
timharrison-moj, line 17: There is still HTML markup `<br/>` in the utils functions. I think markup should be limited to the presentation layer, though I know it isn't in every service. However, converting an address to an array of lines and passing this array of lines into the nunjuncks template gives an opportunity to present it there however each form wants to do it.

timharrison-moj (approved):
LGTM.  Good to go, but I'll think about a refactor into an address macro (reusable).

PR #36, opened 2024-09-05
PRC-57: Wrap all controllers with asyncMiddleware

james-baker-moj:
All controllers that return a promise (which is all of them currently) should be wrapped in asyncMiddleware. This is because if they return an unresolved error in a promise it causes node to crash... This means that any unhandled 400 or 500 from a dependency would result in a crash.


timharrison-moj (approved):
LGTM

PR #37, opened 2024-09-06
PRC-57: Support edit DOB from check answers

james-baker-moj:
Discovered several issues with the old data model. For one `Date` is converted back to string when it's persisted in the session. No idea how this worked so well up until now but if you need to parse it back it's a string. Instead, keep the originally entered day, month and year fields to construct a `Date` as needed.

Also, javascript was doing javascript things with the values for the isDobKnown field being true and false. Converted them to be `'Yes' | 'No'` type instead which matches the actual answers.

### DIFF
--- a/server/@types/journeys/index.d.ts
+++ b/server/@types/journeys/index.d.ts
@@ -15,8 +15,10 @@ declare namespace journeys {
   }
 
   export interface DateOfBirth {
-    isKnown: boolean
-    dateOfBirth?: Date
+    isKnown: 'Yes' | 'No'
+    day?: number
+    month?: number
+    year?: number
   }
 
   export interface ManageContactsJourney {
--- a/server/routes/contacts/create/check-answers/createContactCheckAnswersController.test.ts
+++ b/server/routes/contacts/create/check-answers/createContactCheckAnswersController.test.ts
@@ -2,10 +2,11 @@ import type { Express } from 'express'
 import request from 'supertest'
 import { v4 as uuidv4 } from 'uuid'
 import { SessionData } from 'express-session'
+import * as cheerio from 'cheerio'
 import { appWithAllRoutes, user } from '../../../testutils/appSetup'
 import AuditService, { Page } from '../../../../services/auditService'
-import CreateContactJourney = journeys.CreateContactJourney
 import ContactsService from '../../../../services/contactsService'
+import CreateContactJourney = journeys.CreateContactJourney
 
 jest.mock('../../../../services/auditService')
 jest.mock('../../../../services/contactsService')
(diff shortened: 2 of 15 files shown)


server/routes/contacts/create/check-answers/createContactCheckAnswersController.test.ts
```diff
@@ -16,20 +17,24 @@ const contactsService = new ContactsService(null) as jest.Mocked<ContactsService
...
-  dateOfBirth: {
-    isKnown: true,
-    dateOfBirth: new Date('2024-01-01'),
-  },
-}
+let journey: CreateContactJourney
 beforeEach(() => {
+  journey = {
+    id: journeyId,
+    lastTouched: new Date(),
```
james-baker-moj, line 24: I also need to change lastTouched to not be a Date. I guess it's working because the ISO format makes the order correct alphabetically when it gets converted to a string....
To be done in a different PR

server/@types/journeys/index.d.ts
```diff
@@ -15,8 +15,10 @@ declare namespace journeys {
   }
 
   export interface DateOfBirth {
-    isKnown: boolean
-    dateOfBirth?: Date
+    isKnown: 'Yes' | 'No'
```
james-baker-moj, line 18: I wanted this to be an enum but for whatever reason when it compiles to javascript it can't reference it properly as `journeys.MyEnum`. To be investigated/fixed in another PR

server/routes/contacts/create/check-answers/createContactCheckAnswersController.test.ts
timharrison-moj: Not sure I understand this comment? As lastTouched is still a date?

server/routes/contacts/create/enter-dob/createContactEnterDobController.ts
```diff
@@ -9,7 +9,14 @@ export default class CreateContactEnterDobController implements PageHandler {
...
     const { journeyId } = req.params
     const journey = req.session.createContactJourneys[journeyId]
-    res.render('pages/contacts/create/enterDob', { journey })
+    const view = {
+      journey,
+      isKnown: res.locals?.formResponses?.isKnown ?? journey?.dateOfBirth?.isKnown,
+      day: res.locals?.formResponses?.day ?? journey?.dateOfBirth?.day,
+      month: res.locals?.formResponses?.month ?? journey?.dateOfBirth?.month,
+      year: res.locals?.formResponses?.year ?? journey?.dateOfBirth?.year,
+    }
```
timharrison-moj, line 18: Minor only - if we adopt this method (building the viewModel for forms in controllers/handlers) maybe consistency of its name (view, viewModel)?

timharrison-moj (approved):
LGTM. Minor comments/questions only

server/routes/contacts/create/check-answers/createContactCheckAnswersController.test.ts
james-baker-moj: Yes, so one of the fixes I made to was to change dateOfBirth to not be a Date because those don't serialise in the session properly. Last touched is still a date and should be changed but I didn't want to do it in this PR. It does seem to work correctly for now though...

PR #39, opened 2024-09-06
PRC-57: Change lastTouched to be a date string in session

james-baker-moj:
Dates aren't properly serialised in the session when it's persisted to redis so this can lead to issues when comparing the latest journeys. Converted to a string which is an ISO date instead and then create a Date where required.


timharrison-moj (approved):
LGTM

PR #43, opened 2024-09-10
PRC-53: Include prisoner number in manage and create urls

james-baker-moj:
Add prisoner number into the URLs for create and list prisoner contacts. This assists with redirects and keeping the prisoner number available where query params might become lost.  

Added a temporary button to go straight into the create contact journey from list contacts. This can be removed as we progress the search journey.

This is the first part in linking in the manage and create journeys. The next steps are to return to the original journey once the contact is created, add the prisoner profile to create journey and submit the prisoner contact relationship.


integration_tests/e2e/searchPrisoner.cy.ts
```diff
@@ -60,7 +64,10 @@ context('Contacts', () => {
       searchPrisonerPage.prisonerSearchFormField().clear().type(prisonerNumber)
       searchPrisonerPage.prisonerSearchSearchButton().click()
 
-      Page.verifyOnPage(SearchPrisonerPage).clicktoViewContactsList()
+      Page.verifyOnPage(SearchPrisonerPage).clickToViewContactsList()
+
+      const listContactsPage = new ListPrisonerContactsPage('Smith, John')
+      listContactsPage.checkOnPage()
     })
```
timharrison-moj, line 71: I renamed the searchPrisoner test to manageContacts.... and was building up the journey tests there.

server/@types/journeys/index.d.ts
```diff
@@ -41,4 +42,6 @@ declare namespace journeys {
 
   type YesOrNo = 'YES' | 'NO'
   type YesNoOrDoNotKnow = 'YES' | 'NO' | 'DO_NOT_KNOW'
+
+  type PrisonerJourneyParams = { prisonerNumber: string; journeyId: string }
 }
```
timharrison-moj, line 47: Interesting. I was thinking yesterday that we'd need to store the `journeyId` from the previous journey if we switch between `manage` and `create` .. I see you've had the same thought!

james-baker-moj: Well this is more for the path params as I was copying `{ prisonerNumber: string; journeyId: string }` into dozens of controller methods but yes, we probably need to share the journey id.

server/views/pages/contacts/manage/prisonerSearchResults.njk
```diff
@@ -84,7 +84,7 @@
             {% set rows = [] %}
             {% for result in results.content %}
                 {% set prisonerNameHtml %}
-                   <a href="../list/{{journey.id}}?prisoner={{ result.prisonerNumber }}" class="govuk-link govuk-link--no-visited-state bapv-result-row">
+                   <a href="/prisoner/{{ result.prisonerNumber }}/contacts/list/{{journey.id}}" class="govuk-link govuk-link--no-visited-state bapv-result-row">
                       {{ (result.lastName + ", " + result.firstName) | convertToTitleCase }}
```
timharrison-moj, line 88: Using the prisoner search result here, but (like in other places) we could avoid recalling on every page and use the stored journey values? Fine for now, as mini profile also uses result at present.

timharrison-moj (approved):
LGTM.

server/views/pages/contacts/manage/prisonerSearchResults.njk
james-baker-moj: This is the list of results so think it's still not in a prisoner context here but I get your meaning.

PR #45, opened 2024-09-11
PRC-55: link manage contacts and create new contact journeys

james-baker-moj:
Added a return point to the "create contact" journey and updated the cancel buttons and breadcrumbs to use it.

Successfully creating a contact returns you to the prisoner contact list.

It is currently optional as a convenience for testing as you can currently only get to prisoner contact page by searching. The default is to cancel > home page and show the temporary success page.

Also sets the relationship with defaulted values for now so that the contact appears in the list.

### DIFF
--- a/server/routes/contacts/create/check-answers/createContactCheckAnswersController.test.ts
+++ b/server/routes/contacts/create/check-answers/createContactCheckAnswersController.test.ts
@@ -25,6 +25,7 @@ beforeEach(() => {
     lastTouched: new Date().toISOString(),
     prisonerNumber,
     isCheckingAnswers: false,
+    returnPoint: { type: 'MANAGE_PRISONER_CONTACTS', url: '/foo-bar' },
     names: {
       lastName: 'last',
       firstName: 'first',
@@ -65,16 +66,12 @@ describe('GET /prisoner/:prisonerNumber/contacts/create/check-answers/:journeyId
 
     // Then
     expect(response.status).toEqual(200)
-    expect(response.text).toContain('Contacts')
-    expect(response.text).toContain('Hmpps Contacts Ui')
-    expect(response.text).toContain('Check your answers')
-    expect(auditService.logPageView).toHaveBeenCalledWith(Page.CREATE_CONTACT_CHECK_ANSWERS_PAGE, {
-      who: user.username,
-      correlationId: expect.any(String),
-    })
     expect(journey.isCheckingAnswers).toStrictEqual(true)
     const $ = cheerio.load(response.text)
+    expect($('[data-qa=main-heading]').first().text().trim()).toStrictEqual('Check your answers')
+    expect($('[data-qa=cancel-button]').first().attr('href')).toStrictEqual('/foo-bar')
     expect($('.check-answers-dob-value').first().text().trim()).toStrictEqual('1 January 2024')
+    expect($('[data-qa=contact-list-breadcrumb-link]').first().attr('href')).toStrictEqual('/foo-bar')
   })
 
   it('should render check answers page without dob', async () => {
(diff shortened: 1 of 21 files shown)


server/routes/contacts/create/check-answers/createContactCheckAnswersController.test.ts
```diff
@@ -65,16 +66,12 @@ describe('GET /prisoner/:prisonerNumber/contacts/create/check-answers/:journeyId
 
     // Then
     expect(response.status).toEqual(200)
-    expect(response.text).toContain('Contacts')
-    expect(response.text).toContain('Hmpps Contacts Ui')
-    expect(response.text).toContain('Check your answers')
-    expect(auditService.logPageView).toHaveBeenCalledWith(Page.CREATE_CONTACT_CHECK_ANSWERS_PAGE, {
-      who: user.username,
-      correlationId: expect.any(String),
-    })
```
james-baker-moj, line 74: I refactored these tests to separate the checking of page content and checking of the audit message

server/routes/contacts/create/check-answers/createContactCheckAnswersController.ts
```diff
@@ -32,6 +32,11 @@ export default class CreateContactCheckAnswersController implements PageHandler
     await this.contactService
       .createContact(journey, user)
       .then(() => delete req.session.createContactJourneys[journeyId])
-    res.redirect(`/prisoner/${prisonerNumber}/contacts/create/success`)
+
+    if (journey.returnPoint.type === 'MANAGE_PRISONER_CONTACTS') {
+      res.redirect(journey.returnPoint.url)
+    } else {
+      res.redirect(`/prisoner/${prisonerNumber}/contacts/create/success`)
+    }
```
timharrison-moj, line 40: I'll also test the returning from the end of one mini-journey back into the journey it came from and that the original  journey URL including ID is still in the `returnPoint` of the new mini-journey. Seems like a reasonable way to deal with this, but lets see in practice.

server/routes/contacts/create/start/startCreateContactJourneyController.ts
```diff
@@ -3,18 +3,43 @@ import { v4 as uuidv4 } from 'uuid'
...
+    const { returnJourneyType, returnJourneyId } = req.query
+    const { prisonerNumber } = req.params
+    let returnPoint: ReturnPoint
+    if (returnJourneyType === 'MANAGE_PRISONER_CONTACTS') {
+      returnPoint = {
+        type: 'MANAGE_PRISONER_CONTACTS',
+        url: `/prisoner/${prisonerNumber}/contacts/list/${returnJourneyId}`,
+      }
+    } else {
+      returnPoint = { type: 'HOME', url: '/' }
```
timharrison-moj, line 35: Can you think of a way to clearly explain this process of switching between journeys? I wonder if it needs something in the README to let developers (even us, when we forget in 3 months time) know how this works?

timharrison-moj: Maybe even just a sentence or two in the README references this file / section?

server/views/pages/contacts/create/checkAnswers.njk
```diff
@@ -8,25 +8,10 @@
...
-            href: DPS_HOME_PAGE_URL
-          },
-          {
-              text: "Contacts",
-              href: "/"
-          }
-      ]
-      }) }}
-  </div>
-</div>
```
timharrison-moj, line 26: Has this been a decision now? Limit to 2 levels of breadcrumb? Thought we were still waiting on this, but fine either  way..

timharrison-moj (approved):
LGTM. Very good - solving the issue of multiple journeys and where they intersect. Thanks.

server/views/pages/contacts/create/checkAnswers.njk
james-baker-moj: No I made the change to add the "Prisoner Context" breadcrumb on all pages and de-duped it, it's now included in an import. After that is when I heard we might not even have breadcrumbs on these pages. Still, it's better than the broken ones we had before.

server/routes/contacts/create/check-answers/createContactCheckAnswersController.ts
james-baker-moj: I was planning to remove the success page but that had the side effect of breaking every single test so deferred. I think the default should actually be straight to prisoner contacts list page but there is no entry point for that currently.

PR #46, opened 2024-09-12
PRC-55: Contacts with no DOB blow up the list contacts page

james-baker-moj:
Now shows 'Not provided' instead, as per the check answers page in create.


timharrison-moj (approved):
LGTM. Test-driven development in action 👍

PR #47, opened 2024-09-12
PRC-107: Add reference data service and use for titles

james-baker-moj:
Integrated the reference code API. Provide a dedicated service as we will be switching the source to different endpoints in the future. Set up stubbing mechanism for other reference data and integration tests.


timharrison-moj (approved):
LGTM

PR #52, opened 2024-09-17
PRC-119 - Emergency contact & next of kin create journey

james-baker-moj:


integration_tests/e2e/createContacts.cy.ts
```diff
@@ -17,30 +19,19 @@ context('Create Contacts', () => {
...
-    const enterDobPage = new EnterContactDateOfBirthPage('Last, First')
-    enterDobPage.checkOnPage()
-    enterDobPage //
-      .selectIsKnown('NO')
-      .clickContinue()
+    nameIsFirstLast()
+    relationshipIsMother()
+    isEmergencyContact('NO')
+    isNextOfKin('YES')
+    dobIsUnknown()
```
timharrison-moj, line 35: Bit uncertain about these shorthand functions? They don't really show the journey happening between pages, so the user flow is a little obscured by what look like page validation functions only? Could the specific pages stay visible, but using chained functions demonstrate a flow? Just a thought - something like: 
`
    const relationShipPage = new EnterDobPage()
        .checkOnPage()
        .selectIsKnown('NO'
        .clickContinue()
   
   const emergencyContactPage = relationShipPage
      .checkOnPage()
      .verifyListContains('MOT')
      .select('MOT')
      .continue()
`
etc. ..

timharrison-moj (approved):
LGTM. All looking very neat.
One question is around the chaining of cypress tests to retain the flow between pages -  think BVLS UI does this pretty well, as an example. Possibly also A&A UI and CVL UI too. Could be something to think about - doesn't hold up this PR from going in though.

PR #54, opened 2024-09-18
PRC-53: Standardise formatting of name

james-baker-moj:
Add a nunjucks filter for formatting objects with names to be used with prisoners and contacts. Prisoners currently do not show middle names while contacts do as per relevant tickets.

Capitalisation and middle names were missing from create journey so those have been fixed as a result.


timharrison-moj (approved):
LGTM. And a helpful template for me to look at an address formatting util/macro.

PR #61, opened 2024-09-23
PRC-46 Contact Search Table

rosariodiaferia-moj:
**Ticket**: https://dsdmoj.atlassian.net/browse/PRC-46
**Description**: enable the user to search for a contact if the one they were looking for in the prisoner contact list is not found.

**Screehshot**:

Hmpps-Contacts-Ui-Search-for-Contact-09-23-2024_02_56_PM


assets/scss/local.scss
```diff
@@ -7,7 +7,7 @@
   padding: 30px 15px;
 }
 
-#active-contacts, #inactive-contacts {
+#active-contacts, #inactive-contacts, #contact-search-results {
```
james-baker-moj, line 10: should it just have a class like `.contact-list` rather than all the ids?

server/data/contactsApiClient.ts
```diff
@@ -43,4 +45,25 @@ export default class ContactsApiClient extends RestClient {
       user,
     )
   }
+
+  async searchContact(
+    contactSearchRequest: ContactSearchRequest,
+    user: Express.User,
+    pagination?: PaginationRequest,
+  ): Promise<Contact> {
+    const paginationParameters = pagination ?? { page: 0, size: config.apis.prisonerSearchApi.pageSize || 20 }
```
james-baker-moj, line 54: Is it right to use `config.apis.prisonerSearchApi.pageSize` for this? Also, the ticket says 10....

server/utils/utils.ts
```diff
@@ -71,3 +71,27 @@ export const extractPrisonerNumber = (search: string): string | false => {
...
+export const formatDateForApi = (dateOfBirth: string) => {
+  const date = JSON.parse(dateOfBirth)
+  if (date.year && date.month && date.day) {
+    return `${date.year}-${date.month}-${date.day}`
+  }
+  return null
+}
+
+const isListed = (val: string, name: string) => {
+  return val === name || val.toLocaleUpperCase() === name || val.toLocaleLowerCase() === name
```
james-baker-moj, line 84: couldn't you convert both to upper/lower case and do it in one condition `val.toLocalUpperCase() === name.toLocalUpperCase()`

server/views/pages/contacts/manage/contactSearch.njk
```diff
@@ -40,96 +43,184 @@
...
-                    {{ govukButton({
-                        html: "Search",
-                        type: "submit",
-                        classes: 'govuk-!-margin-top-6',
-                        attributes: {"data-qa": "search-button"}
-                    }) }}
-                </div>
-            </form>
+        <div class="govuk-grid-column-full govuk-!-margin-top-4">
+            {% if not view.isContactListed %}
```
james-baker-moj, line 217: So we hide the link if any contact first, last or middle name is an exact match  (case-insensitive) ? That doesn't seem right.

james-baker-moj: >CTA  “The contact is not listed” - takes user to “Contact confirmation” page (covered in PRC-47: Contact confirmation
>CTA will not be displayed until search is performed (this is effectively route into a new journey “Create new contact”)

I think the link is supposed to be displayed once any search is done. With that said, the AC is conflicting as it says it takes you to contact confirmation (wrong, that's for linking a contact) and then "Create new contact" (right, because we don't want to select a contact and instead create a new one)

assets/scss/local.scss
rosariodiaferia-moj: contact-list class added

server/data/contactsApiClient.ts
rosariodiaferia-moj: PaginationRequest  removed and `import Pageable = contactsApiClientTypes.Pageable` added

server/utils/utils.ts
rosariodiaferia-moj: Code refactored to use one condition (`val.toLocalUpperCase() === name.toLocalUpperCase()`)

server/views/pages/contacts/manage/contactSearch.njk
rosariodiaferia-moj: Changed as per AC

rosariodiaferia-moj: Changes added as per AC

server/data/contactsApiClient.ts
```diff
@@ -43,4 +45,25 @@ export default class ContactsApiClient extends RestClient {
       user,
     )
   }
+
+  async searchContact(
+    contactSearchRequest: ContactSearchRequest,
+    user: Express.User,
+    pagination?: Pageable,
+  ): Promise<Contact> {
+    const paginationParameters = pagination ?? { page: 0, size: config.apis.prisonerSearchApi.pageSize || 10 }
```
james-baker-moj, line 54: I think this variable `config.apis.prisonerSearchApi.pageSize` is set to 20.  Shoulduse a different variable for contact list page size or just make it a constant

james-baker-moj (changes requested):
One small thing left

server/data/contactsApiClient.ts
rosariodiaferia-moj: `pageSize` added to `config.apis.contactsApi`

james-baker-moj (approved):
LGTM

PR #74, opened 2024-10-02
PRC-147: Use city, county and country description instead of code

james-baker-moj:
Also, semi-fixed the types for contact search but Page is not currently defined on the back end.


timharrison-moj (approved):
LGTM

PR #75, opened 2024-10-02
PRC-153 - Fix date format for contact search api date of birth (eg. 1990-01-01)

rosariodiaferia-moj:
**Ticket**: https://dsdmoj.atlassian.net/browse/PRC-153

**Description**: Changes added on the util to fix the 400 error - surname and date with correct format for api when day or month starts with zero


james-baker-moj (approved):
LGTM

PR #77, opened 2024-10-03
PRC-152: Re-import contacts types and fix test data factories

james-baker-moj:
Contact search now uses page types with contact search items.

Get contact by id and search contacts were using the same test util method but the types have drifted significantly so those have been split.


timharrison-moj (approved):
LGTM

PR #82, opened 2024-10-09
PRC 149 Contact Confirmation Api Integration

rosariodiaferia-moj:
**Ticket**: https://dsdmoj.atlassian.net/browse/PRC-149

**Description**: API integration for contact Id information page

Hmpps-Contacts-Ui-Contact-Confirmation-10-09-2024_02_32_PM


server/views/pages/contacts/manage/contactConfirmation/basicDetails.njk
```diff
@@ -11,26 +34,23 @@
             text: "Name"
         },
         value: {
-            html: '',
-            attributes: {"data-qa": "confirm-name-value"}
+            html: '<span data-qa="confirm-name-value">' + contactFullNameRow | join(' ') + '</span>'
```
james-baker-moj, line 37: What is the span adding here vs. using the attributes? I see it in a lot of places but it doesn't have any styling associated with it as far as I can tell

server/routes/contacts/manage/contact-confirmation/contactConfirmationController.ts
```diff
@@ -4,19 +4,33 @@ import { Page } from '../../../../services/auditService'
...
+    const { prisonerDetails, user } = res.locals
     const journey = req.session.addContactJourneys[journeyId]
+    const validationErrors = res.locals.validationErrors?.isContactConfirmed
+    let contact: Contact
+
+    try {
+      contact = validationErrors ? null : await this.contactsService.getContact(journey.contactId, user)
+    } catch (error) {
+      logger.error(JSON.stringify(error))
+    }
```
james-baker-moj, line 30: You don't need to catch errors here as the controller is wrapped with `asyncMiddleware` it should handle it correctly if the Promise is rejected.

I am also a bit confused about the logic. Doesn't this mean that if you submit the page without selecting the contact you will get validation errors but not populate the contact in the view model?

rosariodiaferia-moj: Yes, you're right without selecting the contact the service was not executed and the page displayed without info (see video as well). Try/catch removed.

server/views/pages/contacts/manage/contactConfirmation/basicDetails.njk
rosariodiaferia-moj: Cypress does not find `attributes` added on `govukSummaryList` and fails. This is why I have added a span

I noticed that on `checkAnswers.njk` `classes: 'check-answers-dob-value'` is added for `govukSummaryList` which seems to be fine for cypress. 

I can try to change it to classes?

rosariodiaferia-moj: replaced with classes attribute

james-baker-moj (approved):
LGTM

PR #99, opened 2024-10-17
PRC-209: Increase redis TTL to same as token validity (12 hours)

james-baker-moj:
This should ensure session data is not lost unless re-authenticating.

Also, made the integration tests fire up redis as well.


timharrison-moj (approved):
LGTM

timharrison-moj (approved):
LGTM

PR #112, opened 2024-10-31
PRC-169 : Manage contact staff status

prabash-moj:
PRC-169 : Manage contact staff status


server/routes/contacts/manage/staff/manageContactStaffController.test.ts
```diff
@@ -0,0 +1,74 @@
...
+    const $ = cheerio.load(response.text)
+
+    // Then
+    expect(response.status).toEqual(200)
+    expect($('#isStaff').has('option')).toBeDefined()
+    expect(auditService.logPageView).toHaveBeenCalledWith(Page.MANAGE_CONTACT_UPDATE_STAFF_PAGE, {
+      who: user.username,
+      correlationId: expect.any(String),
+    })
+  })
```
james-baker-moj, line 55: Should be a test here or in cypress that the correct option is filled

server/views/pages/contacts/manage/contactConfirmation/staffStatus.njk
```diff
@@ -0,0 +1,28 @@
...
+        title: {
+            text: "Staff details"
+        }
+    },
+    rows: [{
+        key: {
+            text: "Staff?"
+        },
+        value: {
+            text: contact.isStaff if contact.isStaff else 'Not provided',
```
james-baker-moj, line 14: There seem to be two places with change links which can't be right. This one looks wrong on the else case. Should be "No" but I don't think it needs to be part of contactConfirmation at all....

james-baker-moj: Looking at the ticket I don't see anything related to adding this on contact confirmation. Was it in another JIRA? It definitely shouldn't have an edit link and even if it did I think your back link on the update page wouldn't work properly.

prabash-moj: oh my bad, was following a bad example, 

On a side note: I'm not sure if [manage/contactConfirmation/staffStatus.njk](https://github.com/ministryofjustice/hmpps-contacts-ui/pull/112/files/8e06770f33137fe606a1773716da33feb634fe0c#diff-4799239c24dd546b6f17d188e6362c695c03563f055c13b58d76ae70d970cb07) folder matches the functionality? This is used in ADD flow of the journey, but code is under the Manage folder, Maybe its just me confused 🤷

james-baker-moj (approved):
LGTM

PR #116, opened 2024-11-04
PRC-248: Update types and test data to match API changes for migration

timharrison-moj:
Updated API types from swagger and adjusted test data for removed/added fields.


james-baker-moj (approved):
LGTM

PR #121, opened 2024-11-07
PRC-210: Minor fixes to phone numbers

james-baker-moj:
Show extension in list phone numbers summary card.

Change order of errors to be in field order.

Change button to "Confirm and save"


timharrison-moj (approved):
LGTM

PR #122, opened 2024-11-07
PRC-214 : Create/edit/delete contact identity number - fix order of the error messages


prabash-moj:
PRC-214 : Create/edit/delete contact identity number - fix order of the error messages


james-baker-moj (approved):
LGTM

PR #123, opened 2024-11-08
PRC-168: Sort domestic status using default order

james-baker-moj:
Use the default order provided by the back end which is displaySequence, description.


timharrison-moj (approved):
LGTM

PR #126, opened 2024-11-12
PRC-258: Prevent access prisoner not in caseload

james-baker-moj:
This is handled in the prisoner details middleware rather than in the service as we may need to access limited prisoner data outside of caseload to show related prisoners to contact.


server/views/pages/errors/notFound.njk
```diff
@@ -0,0 +1,17 @@
...
+{% set title =  'Page not found' %}
+
+{% block content %}
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <h1 class="govuk-heading-xl">{{ title }}</h1>
+      <p class="govuk-body">If you typed the web address, check it is correct.</p>
+      <p class="govuk-body">If you pasted the web address, check you copied the entire address.</p>
+      <p class="govuk-body">If the web address is correct or you selected a link or button, contact the DPS helpdesk on 0800 917 5148 or #6598 from inside an establishment.</p>
+      <p class="govuk-body"><a class="govuk-link--no-visited-state" href="{{DPS_HOME_PAGE_URL}}">Go to digital services home page</a></p>
```
timharrison-moj, line 13: I think the content probably wants to be more informative - but one for the designers. It could, for example, explain that the prisoner might have moved, or be outside of your currently active caseload. Not a blocker.

timharrison-moj (approved):
LGTM. Just needs a way to audit the view attempt.

PR #136, opened 2024-11-15
PRC-309: Show relationship details on manage contact

james-baker-moj:


server/@types/contactsApi/index.d.ts
```diff
@@ -4325,6 +4325,39 @@ export interface components {
       property?: string
       ignoreCase?: boolean
     }
+    /** @description Describes the prisoner contact relationship */
+    PrisonerContactRelationshipDetails: {
+      /**
```
timharrison-moj, line 4330: Are we limiting to currentTerm here? I.e they relate to the current (or latest) term in prison?

timharrison-moj (approved):
LGTM

server/@types/contactsApi/index.d.ts
james-baker-moj: It's referenced by id so I guess you could load it but they shouldn't be returned in the contact list right?

PR #141, opened 2024-11-18
PRC-284 H1 name order update

james-baker-moj:
Updated all H1 from Last, First Middle to First Middle Last for contacts or First Last for prisoners. Mini profile and CYA remain reverse order. I also renamed the nunjucks filters as reverseNameFormat actually had first name, first.


prabash-moj (approved):

timharrison-moj (approved):
LGTM

PR #143, opened 2024-11-18
PRC-252: Confirm delete identity number

james-baker-moj:
Plus fix build

### DIFF
--- a/server/routes/contacts/manage/identities/delete/manageContactDeleteIdentityController.ts
+++ b/server/routes/contacts/manage/identities/delete/manageContactDeleteIdentityController.ts
@@ -15,7 +15,6 @@ export default class ManageContactDeleteIdentityController implements PageHandle
     res: Response,
   ): Promise<void> => {
     const { user } = res.locals
-    const { journey } = res.locals
     const { contactId, contactIdentityId } = req.params
     const contactIdNumber = parseInt(contactId, 10)
     const contactIdentityNumber = parseInt(contactIdentityId, 10)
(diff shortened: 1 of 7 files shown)


server/routes/contacts/manage/identities/delete/manageContactDeleteIdentityController.ts
```diff
@@ -29,6 +28,19 @@ export default class ManageContactDeleteIdentityController implements PageHandle
       )
     }
 
+    res.render('pages/contacts/manage/confirmDeleteIdentity', { identityNumber })
+  }
+
+  POST = async (
```
prabash-moj, line 34: Just wondering can this be named as DELTE? so it matches the delete operation?

server/routes/contacts/manage/manageContactsRoutes.ts
```diff
@@ -247,6 +247,11 @@ const ManageContactsRoutes = (
     logPageViewMiddleware(auditService, manageContactDeleteIdentityController),
     asyncMiddleware(manageContactDeleteIdentityController.GET),
   )
+  router.post(
```
prabash-moj, line 250: can we not use   'router.delete(' here?

prabash-moj (approved):

server/routes/contacts/manage/identities/delete/manageContactDeleteIdentityController.ts
james-baker-moj: well it's a form POST so it matches that instead. All of the controllers just have `GET` and `POST` as part of the interface.

PR #148, opened 2024-11-19
PRC-306: Tidy up

james-baker-moj:
Remove commented line and add missing no visit state on contact list links


timharrison-moj (approved):
LGTM

PR #152, opened 2024-11-21
PRC-328: Update prisoner-contact URLs

james-baker-moj:
Also discovered that some manage relationship controllers were not returning to journey returnPoint url so fixed them.


timharrison-moj (approved):
LGTM

PR #154, opened 2024-11-25
PRC-330: Prevent double click

james-baker-moj:
Added the preventDoubleClick option to all buttons. This is an optional javascript feature with fallback to prevent double clicking submit buttons. I was able to reproduce issues fairly consistently by double clicking on create contact or email for example


timharrison-moj (approved):
LGTM

PR #156, opened 2024-11-25
PRC-331 : Sort email addresses alphabetically

prabash-moj:
PRC-331 : Sort email addresses alphabetically (both manage contacts and contact match pages)


james-baker-moj (approved):
LGTM

PR #160, opened 2024-11-27
PRC-334 : Tech - switch language to use reference_codes

prabash-moj:
PRC-334 : Tech - switch language to use reference_codes instead of custom language tables


james-baker-moj (approved):
LGTM

PR #161, opened 2024-11-29
PRC-74: Create restrictions journeys

james-baker-moj:
Journeys to create global and prisoner contact restrictions. Discovered a bug in constructing cypress pages as well which caused some test failures when fixed which is why those tests also changed.


timharrison-moj (approved):
LGTM. Very thorough.

PR #164, opened 2024-12-04
PRC-74: Manage restrictions journeys

james-baker-moj:
Include update global and prisoner-contact plus the success banner.


integration_tests/e2e/updatePrisonerContactRestriction.cy.ts
```diff
@@ -0,0 +1,139 @@
...
+    cy.task('stubGetPrisonerContactRestrictions', {
+      prisonerContactId,
+      response: { prisonerContactRestrictions: [prisonerContactRestriction] },
+    })
+    cy.signIn()
+
+    // TODO visit here from the prisoner contact page instead of directly
+    cy.visit(
+      `/prisoner/${prisonerNumber}/contacts/${contactId}/relationship/${prisonerContactId}/restriction/update/PRISONER_CONTACT/enter-restriction/${restrictionId}?returnUrl=/prisoner/${prisonerNumber}/contacts/manage/${contactId}/relationship/${prisonerContactId}`,
+    )
```
timharrison-moj, line 50: Is this the part that is waiting for Prabash's PR?

server/routes/restrictions/enter-restriction/enterNewRestrictionController.ts
```diff
@@ -7,8 +7,9 @@ import { Navigation } from '../../contacts/common/navigation'
 import ReferenceCode = contactsApiClientTypes.ReferenceCode
 import RestrictionClass = journeys.RestrictionClass
 import { maxLengthForRestrictionClass, RestrictionSchemaType } from '../schema/restrictionSchema'
+import { formatNameFirstNameFirst } from '../../../utils/formatName'
```
timharrison-moj, line 11: 2 x NameFirst?

timharrison-moj: Could be legit - just reads oddly!

timharrison-moj (approved):
LGTM. Nice work, as usual.

server/routes/restrictions/enter-restriction/enterNewRestrictionController.ts
james-baker-moj: Yeah open to suggestions on those method names. It was called reverseNameFormat before but had the first name first... We now have `formatNameFirstNameFirst` and `formatNameLastNameFirst`

integration_tests/e2e/updatePrisonerContactRestriction.cy.ts
james-baker-moj: Yeah I also had to implement the GET for prisoner-contact restrictions which I know he'll be doing as well but that let me test and check it in, shouldn't be too bad. 
It's also supposed to return to the restriction tab after an update so that'll need figuring out.

PR #166, opened 2024-12-06
PRC-323: Add address select type

james-baker-moj:
Initial part of the add address journey to select address type.


server/routes/contacts/manage/addresses/address-type/addressTypeController.ts
```diff
@@ -0,0 +1,75 @@
...
+    )
+  }
+
+  private getSelectedOptions(
+    options: ReferenceCode[],
+    selected?: string,
+  ): Array<{
+    value: string
+    text: string
+    selected?: boolean
```
timharrison-moj, line 65: Although in NOMIS addressType is nullable, do we definitely want to force a selection in our journeys?

server/views/pages/contacts/manage/address/addressType.njk
```diff
@@ -0,0 +1,51 @@
...
+        <div class="govuk-grid-column-three-quarters">
+            <form method='POST'>
+                <input type="hidden" name="_csrf" value="{{ csrfToken }}" />
+                {% set defaultOptions = [
+                    { divider: 'or' },
+                    {
+                        value: "DO_NOT_KNOW",
+                        text: "I don't know",
+                        checked: journey.addressType === 'DO_NOT_KNOW'
+                    }
```
timharrison-moj, line 27: Ah - there is an I do not know option..

timharrison-moj (approved):
LGTM

server/views/pages/contacts/manage/address/addressType.njk
james-baker-moj: Yeah, that's just to make them choose a real option or DO_NOT_KNOW. It'll translate to null when we call the API

PR #177, opened 2024-12-13
PRC-327: Edit address

james-baker-moj:
Also made it so an address is shown even if expired on the manage contact details page so you always have a way to get to view all addresses even if the only address is expired.


prabash-moj (approved):

timharrison-moj (approved):
LGTM.

PR #178, opened 2024-12-16
PRC-318: Add edit delete address phones

james-baker-moj:
Also fixing issues with the address phone display when no extension. Now uses proper summary card rows instead of multi line.


server/routes/contacts/manage/addresses/manageAddressesController.ts
```diff
@@ -28,6 +28,14 @@ export default class ManageAddressesController implements PageHandler {
       address,
       cardTitle: this.getExpiredAddressTitle(address.endDate, address.addressTypeDescription ?? 'Address'),
       mostRelevantAddressLabel: getLabelForAddress(address),
+      cardActions: [
```
prabash-moj, line 31: should'nt this go in the nunjuck template?

prabash-moj (approved):

server/routes/contacts/manage/addresses/manageAddressesController.ts
james-baker-moj: yeah ideally it would but the way this page and the macro were built made it hard to do that in this case. I plan to refactor some more in the future.

PR #189, opened 2024-12-20
PRC-341: Prisoner image default location (esbuild has changed this)

timharrison-moj:
Turns out we needed additional config in circle to account for the different location of images during the build.
Hopefully this resolves - works in unit tests, int tests, running locally and in circle CI now.
*should* also work in DEV, PRE, PROD etc..

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -52,7 +52,6 @@ jobs:
             - node_modules
             - build
             - dist
-            - assets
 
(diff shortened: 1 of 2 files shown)


james-baker-moj (approved):
LGTM - I have always found it weird why this isn't a common service we can just put an absolute url to. seems odd that every service re-implements this

PR #190, opened 2025-01-06
PRC-165: Fix age calculation

james-baker-moj:
Some tests had started failing as date-fns rounds to the nearest year in minutes which doesn't work properly for age due to leap years. Changed to a custom implementation which diffs the years and then takes off one year if it hasn't been their birthday yet this year. If you were born on 29th Feb then your age doesn't tick up until 1st March on non-leap years which is safest for an under-18 check. Renamed function to reflect it's only used for age.


timharrison-moj (approved):
LGTM

PR #192, opened 2025-01-06
PRC-355 : add validation to ensure the start date is before the expir…

prabash-moj:
PRC-355 :  add validation to ensure the start date is before expiry date.

### DIFF
--- a/server/routes/restrictions/schema/restrictionSchema.ts
+++ b/server/routes/restrictions/schema/restrictionSchema.ts
@@ -1,6 +1,6 @@
 import z from 'zod'
 import { Request } from 'express'
-import { isValid, parse } from 'date-fns'
+import { format, isValid, parse } from 'date-fns'
 import { createSchema, SchemaFactory } from '../../../middleware/validationMiddleware'
 import RestrictionClass = journeys.RestrictionClass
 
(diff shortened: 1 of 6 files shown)


server/routes/restrictions/schema/restrictionSchema.ts
```diff
@@ -42,6 +42,24 @@ export const restrictionSchema =
         .max(maxCommentLength, `Comment must be ${maxCommentLength} characters or less`)
         .optional()
         .transform(value => value?.trim() || undefined),
+    }).superRefine((data: { startDate: string; expiryDate: string }, ctx) => {
+      const parseDate = (date: string) => parse(date, 'dd/MM/yyyy', new Date())
+      const isDateValid = (date: string) => date !== undefined && isValid(parseDate(date))
+
```
james-baker-moj, line 48: You are parsing the date twice here, I think it could be simplified. You don't need to return anything from `superRefine` so you could parse date once and then have a single if for whether to add the issue or not? Something like `if isValid(start) && isValid(expiry) && start > expiry` ?

prabash-moj: hmm, I think the tricky bit here is the undefined check causes it to make multiple parses, after considering all the different scenarios for date values, I was only able to simplify this very little, good point about returning void so I have removed the return true/false values from  this method.

james-baker-moj (approved):
LGTM

PR #193, opened 2025-01-09
PRC-402: Remove estimated dob question

james-baker-moj:
This is not required or supported in NOMIS so is being removed. Also means the update DOB page is now a bounce page rather than mini-journey.


timharrison-moj (approved):
LGTM.

PR #195, opened 2025-01-13
PRC-399 : Tech ticket - Add support for creating official contacts in…

prabash-moj:
[PRC-399](https://dsdmoj.atlassian.net/browse/PRC-399) : Tech ticket - Add support for creating official contacts in the API

### Create contact scenario :
Create relationship
image

Update relationship
image

Updated relationship
image

### Add existing contact 
/ add and update relationship
image

[PRC-399]: https://dsdmoj.atlassian.net/browse/PRC-399?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


server/services/contactsService.ts
```diff
@@ -61,7 +62,7 @@ export default class ContactsService {
       contactId: journey.contactId,
       relationship: {
         prisonerNumber: journey.prisonerNumber,
-        relationshipCode: journey.relationship.type,
```
james-baker-moj, line 64: I think this also needs `relationshipType: 'S',`. This is for the add existing contact flow. The other one above is add a new contact.

server/views/partials/linkedPrisonerCard.njk
```diff
@@ -3,7 +3,7 @@
 {% macro linkedPrisonerCard(linkedPrisoner) %}
     {% set relationshipHtml = '<ul>' %}
     {% for relationship in linkedPrisoner.relationships %}
-        {% set   relationshipHtml = relationshipHtml + '<li>' + relationship.contactTypeDescription + ' - ' + relationship.relationshipDescription + '</li>' %}
+        {% set   relationshipHtml = relationshipHtml + '<li>' + relationship.contactTypeDescription + ' - ' + relationship.relationshipToPrisonerDescription + '</li>' %}
```
james-baker-moj, line 6: `relationship.contactTypeDescription` think this might be one we missed in the API. Should it be `relationshipTypeDescription` now?

integration_tests/e2e/addExistingContact.cy.ts
```diff
@@ -133,7 +133,7 @@ context('Add Existing Contact', () => {
         contactId,
         relationship: {
           prisonerNumber: 'A1234BC',
-          relationshipCode: 'MOT',
+          relationshipToPrisoner: 'MOT',
```
james-baker-moj, line 136: I'd expect these add existing contact tests to also specify S or O?

server/views/partials/linkedPrisonerCard.njk
prabash-moj: It matches the field [here](https://github.com/ministryofjustice/hmpps-contacts-ui/blob/8435eb1036c665ad5f2ebacdf5ea3cae6d827f4b/server/%40types/contactsApi/index.d.ts#L5568), and on the API [here](https://github.com/ministryofjustice/hmpps-contacts-api/blob/fb647e75238cb0e552eed71a878651ddd341404b/src/main/kotlin/uk/gov/justice/digital/hmpps/hmppscontactsapi/model/response/LinkedPrisonerRelationshipDetails.kt#L28),

integration_tests/e2e/addExistingContact.cy.ts
prabash-moj: good spot, I think we have an issue with the test as this change did not pick up with the unit or Cypress tests, I will dig bit more, At least Cypress should have failed. also wondering why typescript did not complain about missing field as it would do in other places. 🤷

james-baker-moj (approved):
LGTM

PR #196, opened 2025-01-15
PRC-399 : Tech ticket - rename to relationship type description

prabash-moj:
PRC-399 : Tech ticket - rename to relationship type description


james-baker-moj (approved):
LGTM

PR #201, opened 2025-01-20
PRC-382: Add relationship type to CYA

james-baker-moj:
Added relationship type to CYA for new and existing contact linking.

Changing the relationship type requires re-entry of the relationship to prisoner.


timharrison-moj (approved):
LGTM

PR #226, opened 2025-01-22
PRC-240 Revamp link existing contact CYA page

bill-moj:
PRC-240

image

removed "Approved for visits" since that page is not available yet
image


.circleci/config.yml
```diff
@@ -18,7 +18,7 @@ parameters:
 
   node-version:
     type: string
-    default: 20.18-browsers
+    default: 22.10-browsers
```
james-baker-moj, line 21: Does this match the runtime version of our envs? I thought it was on 20.18-ish.

server/routes/contacts/add/check-answers/createContactCheckAnswersController.ts
```diff
@@ -58,7 +70,44 @@ export default class CreateContactCheckAnswersController implements PageHandler
       formattedFullName,
       navigation: navigationForAddContactJourney(this.PAGE_NAME, journey),
     }
-    res.render('pages/contacts/add/checkAnswers', view)
+    res.render('pages/contacts/add/new/checkAnswers', view)
+  }
+
+  private getForExistingContact = async (
```
james-baker-moj, line 76: this looks largely duplicated between the two methods. Does it need to be? It looks like the only difference is the template and that this one doesn't generate a name... This contains a small amount of logic so at very least should be put into a shared method

```
   const relationshipToPrisonerDescription = await this.referenceDataService.getReferenceDescriptionForCode(
      **journey.relationship.relationshipType === 'S'
        ? ReferenceCodeType.SOCIAL_RELATIONSHIP
        : ReferenceCodeType.OFFICIAL_RELATIONSHIP,**
      journey.relationship.relationshipToPrisoner,
      user,
    )
```

.circleci/config.yml
bill-moj: i've updated it to 22 in package.json and in the dockerfile. this is the last bit i missed updating.

james-baker-moj (approved):
LGTM

PR #227, opened 2025-01-24
Add typecheck rules to non-test files

bill-moj:


server/@types/journeys/index.d.ts
```diff
@@ -124,7 +124,7 @@ declare namespace journeys {
     town?: string
     county?: string
     postcode?: string
-    country: string
+    country?: string
```
james-baker-moj, line 127: Country is mandatory on addresses. Maybe the form is also wrong?

server/routes/contacts/add/enter-dob/createContactEnterDobController.ts
```diff
@@ -10,24 +10,24 @@ import captionForAddContactJourney from '../addContactsUtils'
...
       journey,
       caption: captionForAddContactJourney(journey),
-      isKnown: res.locals?.formResponses?.isKnown ?? journey?.dateOfBirth?.isKnown,
-      day: res.locals?.formResponses?.day ?? journey?.dateOfBirth?.day,
-      month: res.locals?.formResponses?.month ?? journey?.dateOfBirth?.month,
-      year: res.locals?.formResponses?.year ?? journey?.dateOfBirth?.year,
+      isKnown: res.locals?.formResponses?.['isKnown'] ?? journey?.dateOfBirth?.isKnown,
+      day: res.locals?.formResponses?.['day'] ?? journey?.dateOfBirth?.day,
+      month: res.locals?.formResponses?.['month'] ?? journey?.dateOfBirth?.month,
+      year: res.locals?.formResponses?.['year'] ?? journey?.dateOfBirth?.year,
```
james-baker-moj, line 22: This is interesting, I guess there is no way to coerce to the Zod form? If we are here then the form was invalid so even then it would be zod form but everything optional?

server/@types/journeys/index.d.ts
bill-moj: in UsePrisonerAddressController, we populate journey.addressLines with prisonerAddressService.getPrimaryAddress:
https://prison-api-dev.prison.service.justice.gov.uk/swagger-ui/index.html#/offenders/getAddressesByOffenderNo

and the country prop is nullable according to prison-api's doc.

james-baker-moj: ah damn, ok I think we should use DEFAULT_COUNTRY in that case as per EnterAddressController?

bill-moj: revert it to be non-nullable and added DEFAULT_COUNTRY to UsePrisonerAddressController

james-baker-moj (approved):
LGTM good work

PR #236, opened 2025-01-30
Integrate Sentry alerts

bill-moj:


server/app.ts
```diff
@@ -61,6 +65,12 @@ export default function createApp(services: Services): express.Application {
   app.use(setUpSuccessNotificationBanner())
   app.use(routes(services))
 
+  app.get('/test-error', (_req, _res) => {
+    throw Error('error to test sentry')
+  })
```
james-baker-moj, line 70: does this stay forever or you just want to test and we can remove?

PR #238, opened 2025-01-31
PRC-239 Edit Employments screen

bill-moj:
PRC-239

image

image


server/routes/contacts/manage/addresses/addressesMiddleware.ts
```diff
@@ -1,24 +1,21 @@
...
-      req.session.addressJourneys = {}
-    }
-    if (!req.session.addressJourneys[journeyId]) {
-      logger.warn(
-        `Address journey (${journeyId}) not found in session for user ${res.locals.user?.username}. Rendering not found.`,
-      )
-      return res.render('pages/errors/notFound')
-    }
-    req.session.addressJourneys[journeyId].lastTouched = new Date().toISOString()
+const ensureInAddressJourney = async (req: Request<PrisonerJourneyParams>, res: Response, next: NextFunction) => {
```
james-baker-moj, line 5: We made them all `asyncMiddleware` as there was some weird behaviour when there were multiple middlewares and some are async and some are not. If the non-async one raises an error then it causes node to exit.

server/@types/journeys/index.d.ts
```diff
@@ -145,6 +161,12 @@ declare namespace journeys {
     contactId?: string
   }
 
+  type JourneyData = Partial<{
+    updateEmployments: UpdateEmploymentsJourney
+  }> & {
+    journeyId: string
+  }
```
james-baker-moj, line 168: This feels less type safe to me than the existing methodology. With the current approach. If there is a journey with id 123 then I know it's of the correct type. With this I could have any type of journey, there is only one currently so it's non-optional but you have to use Partial to achieve that which to me is effectively making it optional.

server/middleware/setUpJourneyData.ts
```diff
@@ -0,0 +1,24 @@
+import { NextFunction, Request, Response } from 'express'
+import { v4 as uuidv4 } from 'uuid'
+import TokenStore from '../data/tokenStore/tokenStore'
+import JourneyData = journeys.JourneyData
+
+const JOURNEY_EXPIRE_TIME_IN_SECONDS = 20 * 60 * 60
+
+export default function setUpJourneyData(store: TokenStore) {
```
james-baker-moj, line 8: While this appears simpler than the other middleware, that's because it does not prevent denial of service attacks as per original session handling methodology where we limit it to 10 of any type within the session and kick out the oldest.

server/@types/journeys/index.d.ts
```diff
@@ -67,6 +69,14 @@ declare namespace journeys {
     inactivateListPage?: number
   }
 
+  export interface UpdateEmploymentsJourney {
+    prisoner: PrisonerDetails
```
james-baker-moj, line 73: PrisonerDetails are already populated in locals with `populatePrisonerDetailsIfInCaseload` middleware, not sure what advantage there is to having them directly on the journey as well.

server/routes/contacts/manage/addresses/addressesMiddleware.ts
bill-moj: putting asyncMiddleware on the last RequestHandler have had worked fine for me (without it i've got those weird error too). it seems to be sufficient to just have it on the last handler but not the middleware.
tests are passing and we have Sentry alert to report if anything goes wrong, so i think it's safe to take them out.

server/middleware/setUpJourneyData.ts
bill-moj: right, my approach here can't prevent one malicious user from taking up all the redis memory.
(while it's possible to do a wildcard search to get a list of keys matching a pattern, that's an expensive operation that should be avoided)
i'll refactor this to follow how the other journeys work then.

server/@types/journeys/index.d.ts
bill-moj: will rework this bit to follow the other journeys' pattern

server/@types/journeys/index.d.ts
bill-moj: populatePrisonerDetailsIfInCaseload would query for prisoner details from prison-api on every GET calls in the journey. i think that's an overkill. we just need to load it once at the start of the journey, put it in the journey data here and reuse it in the subsequent GET calls.

server/routes/contacts/manage/update-employments/updateEmploymentsController.ts
```diff
@@ -0,0 +1,63 @@
...
+      },
+      employments: contact.employments,
+      returnPoint: { url: req.query.returnUrl },
+    }
+    res.redirect(
+      `/prisoner/${prisonerNumber}/contacts/manage/${contactId}/update-employments/${req.journey!.journeyId}`,
+    )
+  }
+
+  ensureInJourney = async (req: Request<PrisonerJourneyParams>, res: Response, next: NextFunction) => {
```
james-baker-moj, line 38: Same with this, any reason it couldn't be in it's own file with it's own tests?

server/utils/sorters.ts
```diff
@@ -1,9 +1,6 @@
...
 
-export const employmentSorter = (
-  a: components['schemas']['EmploymentDetails'],
-  b: components['schemas']['EmploymentDetails'],
-) =>
+export const employmentSorter = (a: EmploymentDetails, b: EmploymentDetails) =>
   Number(b.isActive) - Number(a.isActive) ||
   a.employer.organisationName.localeCompare(b.employer.organisationName) ||
-  b.employmentId - a.employmentId
+  (b.employmentId ?? 0) - (a.employmentId ?? 0)
```
james-baker-moj, line 6: is there a test for whatever caused you do add this?

server/routes/contacts/manage/addresses/addressesMiddleware.ts
james-baker-moj: ok is there an advantage to taking them out though? This way if someone happens to use it last they introduce a bug potentially.

server/@types/journeys/index.d.ts
james-baker-moj: I think that's premature optimisation. The important thing is that we check the caseload on every request. If we want to improve performance we can always implement a cache at a lower level. This encourages you not to use the middleware.  Populating it only using the middleware means we never display prisoner data without checking the caseload.

server/routes/contacts/manage/update-employments/updateEmploymentsController.ts
```diff
@@ -0,0 +1,63 @@
...
+import { Page } from '../../../../services/auditService'
+import logger from '../../../../../logger'
+import PrisonerJourneyParams = journeys.PrisonerJourneyParams
+import { employmentSorter } from '../../../../utils/sorters'
+
+export default class UpdateEmploymentsController implements PageHandler {
+  constructor(private readonly contactsService: ContactsService) {}
+
+  public PAGE_NAME = Page.MANAGE_CONTACT_UPDATE_EMPLOYMENTS_PAGE
+
```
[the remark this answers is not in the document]
bill-moj: this can be split out yes. 
no page is rendered so i don't think we should log a page view event for the start route tho.

server/routes/contacts/manage/update-employments/updateEmploymentsController.ts
bill-moj: this will be reused by other route/controllers in the same journey and should definitely be split out.

server/utils/sorters.ts
bill-moj: the a/c only really mentioned the top two sort logic.
i added a third tie breaker because we don't prevent users from adding employments with the same employer and same active/inactive status, and yet i want the sorting order to be deterministic.

it wouldn't work when the user is in progress adding multiple new employment records with same employer+status tho because there won't be an employmentId for those.

server/routes/contacts/manage/addresses/addressesMiddleware.ts
bill-moj: there shouldn't be a legitimate use case for putting a middleware as the last request handler
the advantage is to reduce the amount of unnecessary boilerplates : )

server/routes/contacts/manage/update-employments/updateEmploymentsController.ts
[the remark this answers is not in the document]
james-baker-moj: ok we are logging them elsewhere with the aim of distinguishing how many journeys are started vs. how many times a page is visited within a journey. e.g., if I change something from CYA then I visit that page more than once but I'm only starting the journey once. With that said, these are for audit team rather than metrics which I misunderstood originally so we could get rid of the start page audits.

james-baker-moj (approved):
LGTM

PR #242, opened 2025-02-04
fix null address start date bug, and

bill-moj:
fix prometheus alert severity label


james-baker-moj (changes requested):
Can you add a test please and name it something like "can edit an address from NOMIS with missing start date" otherwise I'll only end up making it mandatory again accidentally.  Doesn't need to be cypress test as should be reproducible using jest right?

james-baker-moj (approved):
LGTM thanks

PR #252, opened 2025-02-10
PRC-445 : Update UI project to point to personal-relationships-api 

prabash-moj:
PRC-445 : Update UI project to point to personal-relationships-api


timharrison-moj (approved):
LGTM

PR #259, opened 2025-02-11
PRC-236 Add check-employer page

bill-moj:


server/utils/utils.ts
```diff
@@ -110,3 +110,25 @@ export const convertToSortableColumns = (headings: { text: string; key?: string
     }
   })
 }
+
+const isLowerCase = (val: string): boolean => /^[a-z]*$/.test(val)
+
+const lowercaseExceptAcronym = (val: string): string => {
```
james-baker-moj, line 116: Nice, was thinking about needing to lowercase something that had an acronym as one of the options. Will see if this works

james-baker-moj (approved):
LGTM nice job

PR #264, opened 2025-02-12
PRC-237 Add employment-status page

bill-moj:
PRC-237

image


server/routes/contacts/manage/update-employments/employment-status/employmentStatusSchema.ts
```diff
@@ -0,0 +1,10 @@
+import { z } from 'zod'
+import { createSchema } from '../../../../../middleware/validationMiddleware'
+
+const ERROR_MESSAGE = 'Select whether this employment is active or inactive'
+
+export const employmentStatusSchema = createSchema({
+  isActive: z.enum(['true', 'false'], { message: ERROR_MESSAGE }).transform(val => val === 'true'),
```
james-baker-moj, line 7: how come you went for YES/NO in the other schema and true/false in this one? 
I generally prefer the YES/NO to avoid any accidental javascript truthy nonsense...

james-baker-moj (approved):
LGTM one question on schema choice

server/routes/contacts/manage/update-employments/employment-status/employmentStatusSchema.ts
bill-moj: i use true/false so that it's aligned with employment.isActive property.
this schema also handle transforming the string true/false input into proper boolean, so that in the controller req.body.isActive is of the same type as journey.employment[:index].isActive, and doesn't need any further logic / type-casting.

i could use YES/NO as the string value but that's not what the radio labels say, and with true/false it's easier to understand that these values are supposed to be parsed into booleans.

PR #273, opened 2025-02-14
PRC-259 Patch contact employments with success notification

bill-moj:
PRC-259

image

### DIFF
--- a/server/views/pages/contacts/manage/updateEmployments/index.njk
+++ b/server/views/pages/contacts/manage/updateEmployments/index.njk
@@ -105,7 +105,7 @@
             html: "Confirm and save",
             preventDoubleClick: true
           }) }}
-          <a class="govuk-link govuk-link--no-visited-state govuk-!-font-size-19" href="{{ returnPoint.url }}" data-qa="cancel-button">Cancel</a>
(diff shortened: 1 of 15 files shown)


server/views/pages/contacts/manage/updateEmployments/index.njk
```diff
@@ -105,7 +105,7 @@
             html: "Confirm and save",
             preventDoubleClick: true
           }) }}
-          <a class="govuk-link govuk-link--no-visited-state govuk-!-font-size-19" href="{{ returnPoint.url }}" data-qa="cancel-button">Cancel</a>
+          <a class="govuk-link govuk-link--no-visited-state govuk-!-font-size-19" href="{{ returnPoint.url }}#professional-information" data-qa="cancel-button">Cancel</a>
```
james-baker-moj, line 108: Could you not put the anchor in the link into this journey?  If we changed the return point it might not make sense anymore...  Anyway, we can probably hardcode these now after yesterdays discussion so happy to pick that up separately.

james-baker-moj (approved):
LGTM

server/views/pages/contacts/manage/updateEmployments/index.njk
bill-moj: right, we can align the return url strategy for all the various journeys / single-page-bounce-backs later

PR #275, opened 2025-02-14
PRC-279: Update delete contact phone

james-baker-moj:
Some minor text tweaks plus now returns directly to contact details on cancel or success. Made edit contact methods not require a return URL either.

### DIFF
--- a/server/views/pages/contacts/manage/confirmDeletePhone.njk
+++ b/server/views/pages/contacts/manage/confirmDeletePhone.njk
@@ -4,21 +4,30 @@
 {% from "govuk/components/summary-list/macro.njk" import govukSummaryList %}
 
 {% set pageTitle = applicationName + " - Delete contact phone number" %}
-{% set title = "Are you sure you want to delete this phone number?" %}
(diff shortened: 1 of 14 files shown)


server/views/pages/contacts/manage/confirmDeletePhone.njk
```diff
@@ -4,21 +4,30 @@
...
                 <input type="hidden" name="_csrf" value="{{ csrfToken }}" />
                 {{ govukSummaryList({
                     rows: [
+                      {
+                        key: {
+                          text: "Type"
+                        },
+                          value: {
+                            html: phone.phoneTypeDescription,
+                            classes: 'type-value'
```
bill-moj, line 28: text, not html?

also does it need to be in sentence case, eg `Alternative business` rather than `Alternative Business`?

bill-moj (approved):

server/views/pages/contacts/manage/confirmDeletePhone.njk
james-baker-moj: Yeah but we're going to fix that in the reference data rather than the front end.

PR #280, opened 2025-02-21
PRC-265 Contact details-Edit gender

bill-moj:
PRC-265

image

image

### DIFF
--- a/server/views/pages/contacts/manage/contactDetails/manageGender.njk
+++ b/server/views/pages/contacts/manage/contactDetails/manageGender.njk
@@ -14,7 +14,7 @@
 {% include 'partials/formErrorSummary.njk' %}
 <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
-        <span class="govuk-caption-l">Contacts</span>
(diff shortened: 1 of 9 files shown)


server/views/pages/contacts/manage/contactDetails/manageGender.njk
```diff
@@ -24,19 +25,22 @@
...
               },
+              errorMessage: validationErrors | findError('gender'),
               id: "gender",
                 name: "gender",
                 items: genderOptions
             }) }}
             <div class="govuk-button-group">
                 {{ govukButton({
-                    text: "Confirm and save",
+                    text: submitButtonLabel or "Confirm and save",
```
james-baker-moj, line 35: Super minor but we have called this continueButtonLabel everywhere else that it's parameterised. Also does it need the default? We should remove the default if possible.

bill-moj: not really minor imo
keeping the param names consistent will make things easier : )

btw if this pattern of button group with conditional cancel link is going to be repeated for some more pages, we might make a njk macro out of it

james-baker-moj (approved):
LGTM

PR #288, opened 2025-02-24
PRC-276: Refactor relationship to prisoner options

james-baker-moj:
Move group code, hint text and default option logic out of the controllers into a util.

### DIFF
--- a/server/utils/relationshipTypeUtils.ts
+++ b/server/utils/relationshipTypeUtils.ts
@@ -1,0 +1,28 @@
+import ReferenceCodeType from '../enumeration/referenceCodeType'
+
+export const relationshipToPrisonerOptionsForRelationshipType = (
+  relationshipType: string,
+  contactName: string,
+): {
+  groupCodeForRelationshipType: ReferenceCodeType
+  hintText: string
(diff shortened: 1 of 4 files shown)


server/utils/relationshipTypeUtils.ts
```diff
@@ -0,0 +1,28 @@
...
+  let defaultSelectLabel
+  if (relationshipType === 'S') {
+    groupCodeForRelationshipType = ReferenceCodeType.SOCIAL_RELATIONSHIP
+    hintText = `For example, if ${contactName} is the prisoner’s uncle, select ‘Uncle’.`
+    defaultSelectLabel = 'Select social relationship'
+  } else {
+    groupCodeForRelationshipType = ReferenceCodeType.OFFICIAL_RELATIONSHIP
+    hintText = `For example, if ${contactName} is the prisoner’s doctor, select ‘Doctor’.`
+    defaultSelectLabel = 'Select official relationship'
+  }
```
bill-moj, line 22: i think nunjucks macro or filter would be a better place for these sort of content. but it doesn't matter much anyway.

bill-moj (approved):

PR #302, opened 2025-03-03
PRC-462 Combine select language and select interpretation requirements into one page

bill-moj:
PRC-462

image

image


server/views/pages/contacts/manage/contactDetails/manageLanguageAndInterpreter.njk
```diff
@@ -0,0 +1,94 @@
...
+
+             {% for item in spokenLanguages %}
+                {% set languages = (languages.push(
+                    {
+                        value: item.code,
+                        text: item.description,
+                        selected: true if language == item.code else false
+                    }
+                ), languages) %}
+             {% endfor %}
```
james-baker-moj, line 40: I see you've done this in the template rather than controller which is different to most places. 
I've also just got annoyed with those copy/pasted `getSelectOptions` methods and created a util function to do it. Maybe we should combine those and create a nunjucks filter to do it?

james-baker-moj (approved):
LGTM

PR #307, opened 2025-03-04
PRC-460: Add identity documents

james-baker-moj:
Support for adding multiple identity documents at once. Initial implementation of add-another pattern.

Re-wrote the deduplication of error messages to support nested paths.

Error links are all correct and give focus to the correct field.

Complication in the controller to support different actions is only required when javascript is not enabled otherwise the adding/removing  of elements happens inline.

Screenshot 2025-03-04 at 14 45 04

Screenshot 2025-03-04 at 14 44 36

### DIFF
--- a/server/routes/contacts/manage/identities/IdentitySchemas.ts
+++ b/server/routes/contacts/manage/identities/IdentitySchemas.ts
@@ -1,4 +1,5 @@
 import { z } from 'zod'
+import { Request } from 'express'
 import { createSchema } from '../../../../middleware/validationMiddleware'
 import isValidPNC from '../../../../utils/pncValidation'
 
@@ -30,4 +31,21 @@ export const identitySchema = createSchema({
   }
 })
 
+const blankIdentitySchema = createSchema({
+  type: z.string().optional(),
+  identity: z.string().optional(),
+  issuingAuthority: z.string().optional(),
+})
+
+export const identitiesSchema = () => async (request: Request<unknown, unknown, { save?: string }>) => {
+  const childSchema = typeof request.body.save !== 'undefined' ? identitySchema : blankIdentitySchema
+  return createSchema({
+    identities: z.array(childSchema),
+    save: z.string().optional(),
+    add: z.string().optional(),
+    remove: z.string().optional(),
+  })
+}
+
 export type IdentitySchemaType = z.infer<typeof identitySchema>
+export type IdentitiesSchemaType = z.infer<Awaited<ReturnType<ReturnType<typeof identitiesSchema>>>>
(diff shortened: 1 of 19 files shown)


server/routes/contacts/manage/identities/IdentitySchemas.ts
```diff
@@ -30,4 +31,21 @@ export const identitySchema = createSchema({
   }
 })
 
+const blankIdentitySchema = createSchema({
+  type: z.string().optional(),
+  identity: z.string().optional(),
+  issuingAuthority: z.string().optional(),
+})
+
+export const identitiesSchema = () => async (request: Request<unknown, unknown, { save?: string }>) => {
```
bill-moj, line 40: we can probably skip one layer of function call and just do:
```
export const identitiesSchema =  async (request: ...) => {
```

server/views/pages/contacts/manage/addIdentities.njk
```diff
@@ -0,0 +1,116 @@
...
+                      name: "save",
+                      type: "submit",
+                      classes: 'govuk-!-margin-top-6',
+                      attributes: {"data-qa": "continue-button"},
+                      preventDoubleClick: true
+                    }) }}
+                    {% if navigation.cancelButton %}
+                    <a class="govuk-link govuk-link--no-visited-state govuk-!-font-size-19" href="{{ navigation.cancelButton }}" data-qa="cancel-button">Cancel</a>
+                    {% endif %}
+                </div>
```
bill-moj, line 110: not specific to this PR, but we've been copy-and-pasting this bit so many times that i'm tempted to turn it into a njk macro..

server/views/pages/contacts/manage/addIdentities.njk
```diff
@@ -0,0 +1,116 @@
...
+          <span class="govuk-caption-l">Edit identity documentation for a contact</span>
+          <h1 class="govuk-heading-l" data-qa="main-heading">{{ title }}</h1>
+
+          <div data-module="moj-add-another">
+            <form method="POST">
+              <input type="hidden" name="_csrf" value="{{ csrfToken }}" />
+
+              {% for identity in identities %}
+                {% call govukFieldset({
+                  classes: "moj-add-another__item",
```
bill-moj, line 28: nice. i would have just put an h2 here coz i wasn't aware of this component tbf

bill-moj (approved):

server/views/pages/contacts/manage/addIdentities.njk
james-baker-moj: true but this one is actually the only one that's different! 😂 
`name: "save",` is required to make it work without JS.

server/views/pages/contacts/manage/addIdentities.njk
james-baker-moj: copied it from the example for the component. Quite handy actually

PR #311, opened 2025-03-06
PRC-281: Linked prisoners tab

james-baker-moj:
Initial version of linked prisoners tab & table. This will eventually replace the linked prisoner card but it's still used in the link existing prisoner journey.

Currently has no empty state as it'll show all relationships which there must be one in the context of a contact record but may be needed for link existing contact later.

Also fixed the generator script for personal relationship API which I must've broken at some point...

Screenshot 2025-03-06 at 08 13 48

### DIFF
--- a/server/services/contactsService.ts
+++ b/server/services/contactsService.ts
@@ -20,7 +20,6 @@ import CreateContactAddressRequest = contactsApiClientTypes.CreateContactAddress
 import UpdateContactAddressRequest = contactsApiClientTypes.UpdateContactAddressRequest
 import CreateContactAddressPhoneRequest = contactsApiClientTypes.CreateContactAddressPhoneRequest
 import UpdateContactAddressPhoneRequest = contactsApiClientTypes.UpdateContactAddressPhoneRequest
-import LinkedPrisonerDetails = contactsApiClientTypes.LinkedPrisonerDetails
 import PatchEmploymentsRequest = contactsApiClientTypes.PatchEmploymentsRequest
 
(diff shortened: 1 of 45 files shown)


server/services/contactsService.ts
```diff
@@ -20,7 +20,6 @@ import CreateContactAddressRequest = contactsApiClientTypes.CreateContactAddress
 import UpdateContactAddressRequest = contactsApiClientTypes.UpdateContactAddressRequest
 import CreateContactAddressPhoneRequest = contactsApiClientTypes.CreateContactAddressPhoneRequest
 import UpdateContactAddressPhoneRequest = contactsApiClientTypes.UpdateContactAddressPhoneRequest
-import LinkedPrisonerDetails = contactsApiClientTypes.LinkedPrisonerDetails
 import PatchEmploymentsRequest = contactsApiClientTypes.PatchEmploymentsRequest
```
timharrison-moj, line 23: Suggestion only, but these type definitions could all be placed into a single `contact-api-types.ts` file in the @types directory?  It would remove the specific import and type definition in the places they are used.

timharrison-moj (approved):
LGTM

server/services/contactsService.ts
james-baker-moj: We've been moving away from that because `tsc` doesn't correctly identify issues when the import is proxied like that. Both @bill-moj and I have have had a go at fixing it. 
It's not ideal but it has prevented me from making a number of mistakes so far so sticking with it for now.

PR #319, opened 2025-03-07
PRC-469 enter address dates page

bill-moj:
PRC-469


server/routes/contacts/manage/addresses/dates/addressDatesSchemas.ts
```diff
@@ -65,41 +65,41 @@ export const addressDatesSchema = createSchema({
...
-  ) {
-    const formattedFromDate = formatDate(new Date(`${val.fromYear}-${val.fromMonth}-01`), 'MMMM yyyy')
-    ctx.addIssue({
-      code: z.ZodIssueCode.custom,
-      message: `End date must be the same as or after the start date ${formattedFromDate}`,
-      path: ['toDate'],
-    })
-  }
+})
+  .superRefine((val, ctx) => {
```
bill-moj, line 69: the reason to break `transform` into `superRefine` and `transform` is that transform will not be called if the previous validation fails.

by using superRefine instead, we will be able to provide validation errors from this block even when the validations for above failed.

james-baker-moj:
Just looking at the screenshot I think the error messages in the summary are supposed to line up with the field order but it's start/end/start/end (instead of start/start/end/end)

integration_tests/pages/searchContactPage.ts
```diff
@@ -50,7 +50,9 @@ export default class SearchContactPage extends Page {
   }
 
   verifyShowsAddressAs(expected: string): SearchContactPage {
-    this.checkContactSearchTableAddressValue().should('contain.text', expected)
+    const regex = new RegExp(expected.split('<br>').join('<br>\\n?\\s+?'))
+    this.checkContactSearchTableAddressValue().then(element => expect(element.html()).match(regex))
+    // this.checkContactSearchTableAddressValue().should('contain.text', expected)
```
james-baker-moj, line 55: commented out?

server/routes/contacts/manage/addresses/dates/addressDatesSchemas.ts
james-baker-moj: 👍 Normally use superRefine, not sure why I used transform here 🤷

server/views/pages/contacts/manage/contactMethods/address/addressType.njk
```diff
@@ -11,7 +11,7 @@
   {% set hintHtml %}
     <div data-qa="address-reference" class="govuk-body govuk-!-margin-bottom-7 govuk-!-margin-top-2">
       <strong>Address:</strong><br/>
-      {{ formattedAddress | addressToLines | safe }}
+      {{ formattedAddress | addressToLines | escape | nl2br | safe }}
```
james-baker-moj, line 14: do we need a `safenl2br` filter? Kind of annoying needing 4 filters to do one job.

bill-moj: escape nl2br and safe are all builtin nunjucks filters tho
i don't think they are exposed for us to combine

james-baker-moj (approved):
LGTM

PR #324, opened 2025-03-10
PRC-499: Add restrictions cancel page

james-baker-moj:
Add page to confirm cancelling adding a restriction. Also made the middleware return not found instead of returning to start of journey so if you go back you don't get an unexpected blank form (this was only a dev convenience before).

Added some util functions to click buttons or links going to specific pages for more concise cypress tests that more accurately reflect the actual page. (was cancel from cancel instead of "No, return to check answers").

Screenshot 2025-03-10 at 11 18 10


server/views/pages/contacts/restrictions/cancel.njk
```diff
@@ -0,0 +1,53 @@
+{% extends "partials/layout.njk" %}
+{% from "partials/miniProfile/macro.njk" import miniProfile %}
+{% from "govuk/components/button/macro.njk" import govukButton %}
+{% from "govuk/components/summary-list/macro.njk" import govukSummaryList %}
+{% from "govuk/components/warning-text/macro.njk" import govukWarningText %}
+
+{% set pageTitle = "Are you sure you want to cancel adding a restriction? - Manage contact restrictions - DPS" %}
+{% set title = "Are you sure you want to cancel adding a restriction? " %}
```
bill-moj, line 8: mysterious trailing space in the title

integration_tests/pages/page.ts
```diff
@@ -79,6 +87,18 @@ export default abstract class Page {
...
+  clickButtonTo<T>(name: string, constructor: new (...args: any) => T, ...args: any): T {
+    this.clickButton(name)
+    return new constructor(...args)
+  }
+
+  /* eslint-disable @typescript-eslint/no-explicit-any */
+  clickLinkTo<T>(name: string, constructor: new (...args: any) => T, ...args: any): T {
+    this.clickLink(name)
+    return new constructor(...args)
+  }
```
bill-moj, line 100: i like these better than `continueTo` and `cancelTo`!
will start using them in future PRs

bill-moj (approved):

PR #353, opened 2025-03-19
PRC-511: Select language and interpreter

james-baker-moj:
When creating a contact allow select language and interpreter required from additional info page.

Also fixed issue with edit contact details version where the interpreter required value would reset if you didn't select a language.


server/services/contactsService.ts
```diff
@@ -45,9 +45,12 @@ export default class ContactsService {
       lastName: journey.names!.lastName!,
       ...(journey?.names?.middleNames === undefined ? {} : { middleNames: journey.names.middleNames }),
       firstName: journey.names!.firstName!,
-      interpreterRequired: false,
+      interpreterRequired: journey.languageAndInterpreter?.interpreterRequired === 'YES',
```
bill-moj, line 48: a bit odd the question page says `(optional)` in the heading, but really is just a mandatory input that defaults to false

server/services/contactsService.ts
```diff
@@ -45,9 +45,12 @@ export default class ContactsService {
...
       isStaff: journey.isStaff === 'YES',
       ...(journey.gender === undefined ? {} : { genderCode: journey.gender }),
+      ...(journey?.languageAndInterpreter?.language === undefined
+        ? {}
+        : { languageCode: journey?.languageAndInterpreter?.language }),
```
bill-moj, line 53: it's somewhat annoying the `exactOptionalPropertyTypes` rule doesn't treat `var?: type` same as `var: type | undefined` :'(
but it's a small price to pay for the type safety

bill-moj (dismissed):

integration_tests/e2e/selectLanguageAndInterpretationRequirements.cy.ts
```diff
@@ -91,8 +91,8 @@ context('Select Language and interpretation requirements', () => {
     Page.verifyOnPage(EditContactDetailsPage, 'Jones Mason') //
       .clickChangeInterpreterRequiredLink()
 
-    Page.verifyOnPage(SelectLanguageAndInterpreterPage, 'Jones Mason')
-      .continueTo(SelectLanguageAndInterpreterPage, 'Jones Mason')
+    Page.verifyOnPage(SelectLanguageAndInterpreterPage, 'Jones Mason', false)
+      .continueTo(SelectLanguageAndInterpreterPage, 'Jones Mason', false)
       .hasFieldInError('language', 'Select the contact’s first language')
       .hasFieldInError('interpreterRequired', 'Select whether the contact requires an interpreter')
   })
```
bill-moj, line 98:
```suggestion
    Page.verifyOnPage(SelectLanguageAndInterpreterPage, 'Jones Mason', false)
      .continueTo(SelectLanguageAndInterpreterPage, 'Jones Mason', false)
      .hasFieldInError('language', 'Select the contact’s first language')
  })
```

server/services/contactsService.ts
james-baker-moj: It is also language on the same page and you can submit 0, 1 or both fields in the create journey.

server/services/contactsService.ts
james-baker-moj: yeah I was trying to see if there's a way to make the swagger import do the same as this makes it much harder to read. It's way more concise than lots of `if(thing) thing = thing` though

bill-moj (approved):

PR #361, opened 2025-03-25
PRC-235: Contact record update

james-baker-moj:
Add next of kin and emergency contact tags to the contact record. Update the title and include a bottom border.
Add person id.
Fixed the page titles of success pages while I was at it.
Screenshot 2025-03-25 at 07 29 11


timharrison-moj (approved):
LGTM

PR #365, opened 2025-03-26
PRC-377 record employments page to create contact journey

bill-moj:
PRC-244

image

PRC-377

image


server/services/contactsService.ts
```diff
@@ -79,6 +79,12 @@ export default class ContactsService {
         ...(extension === undefined ? {} : { extNumber: extension }),
       }))
     }
+    if (journey.employments?.length && journey.relationship?.relationshipType === 'O') {
```
james-baker-moj, line 82: Can you add a test that it doesn't add employments unless Official please? I think there's already tests for each relationship type so just expand those

james-baker-moj (approved):
LGTM

PR #384, opened 2025-04-01
PRC-550 Revamp contact search page

bill-moj:


server/utils/utils.ts
```diff
@@ -173,3 +176,21 @@ export const referenceCodesToRadiosOrCheckboxes = (
     text: referenceCode.description,
     value: referenceCode.code,
   }))
+
+export const formatDob = (contact: ContactSearchResultItem & { deceasedDate?: string }) => {
+  if (!contact.dateOfBirth) return 'Not provided'
+  const richDate = parseISO(contact.dateOfBirth)
+  if (!isValid(richDate)) return 'Not provided'
+  const age = differenceInYears(new Date(), richDate)
```
james-baker-moj, line 184: I don't think this works correctly. If was used previously but gave incorrect results if you were before or after your birthday.  There is an existing `ageInYears` function which you should adapt/use if you want a filter.  This should also have tests.
I have to do the same thing for contact list and I just used the existing age filter and put the rest in the template. I was planning to update the filter to support the < 1 years old case which is the only thing it doesn't do currently.
```
{{ (item.dateOfBirth | formatDate) if item.dateOfBirth else 'Not provided' }}
{% if item.deceasedDate %}
    <br/><span class="govuk-hint">(Deceased)</span>
{% elseif item.dateOfBirth %}
    <br/><span class="govuk-hint">({{ item.dateOfBirth | ageInYears }} old)</span>
{% endif %}
```

james-baker-moj (changes requested):

server/utils/utils.ts
bill-moj: differenceInYears should work out of the box. i'll update ageInYears util function and reuse it.
don't need the same logic in two places.

server/views/pages/contacts/manage/contactSearch.njk
```diff
@@ -1,164 +1,235 @@
...
+                  key: 'dateOfBirth'
+                },
+                {
+                  text: "Primary or default address"
+                },
+                {
+                  text: "Action",
+                  format: "numeric"
+                }
+              ] | convertToSortableColumns(sort or 'lastName,asc'),
```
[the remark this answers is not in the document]
bill-moj: it only makes a column sortable when it has a `key`, which the address column and action column don't.

server/utils/utils.ts
bill-moj:
```
    ['2024-02-29', '2025-02-28', 'Less than a year'],
    ['2024-02-29', '2025-03-01', '1 year'],
```
weird
the leap year tests for `ageInYears` passed with the `differenceInYears` refactor tho?

james-baker-moj (approved):
LGTM

PR #422, opened 2025-04-23
PRC-653: Don't focus element on error

james-baker-moj:
When we have a deep link to an element, e.g., a change link that focuses the firstName field we use a URL fragment to focus it. This has a side effect of forcing focus to that element even on validation error. As the browser maintains the fragment only on the client side we have no way of knowing if a fragment is in play. To resolve this we reset to a default fragment (#) whenever redirecting on error.


timharrison-moj (approved):
LGTM

PR #423, opened 2025-04-23
PRC-654: UI fixes for contact list

james-baker-moj:
Column heading should be "Visits approval status"
Should include the word "also" when there are both active and inactive restrictions. Add some padding to rows.
Squeeze the name and person id column as small as possible and make sure DOB column does not wrap by setting custom widths.


timharrison-moj (approved):
LGTM

PR #426, opened 2025-04-24
PRC-655: Delete employer link tweak

james-baker-moj:
Make sure the link doesn't wrap and has sufficient padding with long org names.


timharrison-moj (approved):
LGTM

PR #428, opened 2025-04-25
PRC-673: Restrictions table fixes

james-baker-moj:
Remove <hr> between relationship and global restrictions. Add missing middle names for contact.


timharrison-moj (approved):
LGTM

PR #429, opened 2025-04-25
PRC-609: Add links on new contact success back

james-baker-moj:
Links to the contact record and prisoner contact list have been added back for when it's a new contact only.

Existing contacts already provide links back to the contact information.


timharrison-moj (approved):
LGTM

PR #440, opened 2025-05-01
PRC-605 RBAC enforced on add & link contact

james-baker-moj:
Apply RBAC to all controllers under addContactRoutes.


timharrison-moj (approved):
LGTM. A lot of work!

james-baker-moj:
> LGTM. A lot of work!

I'd be lying if I said it wasn't tedious but I think it's worthwhile every controller having GET/POST tests for auth. The bad news is, this is less than half the controllers.

PR #449, opened 2025-05-07
PRC-605: Tweak success page when not authoriser

james-baker-moj:
Remove the word "also" from the hint text when linking an existing contact as we do not show the restrictions content before so "also" doesn't make sense.


timharrison-moj (approved):
LGTM

PR #451, opened 2025-05-07
PRC-723: Remove use of "returnUrl"

james-baker-moj:
This was not actually required any more as all URLs are hardcoded but because it got rendered into links with the templates it was actually an XSS attack vector. This is the easy one to fix.


timharrison-moj (approved):
LGTM

PR #455, opened 2025-05-09
PRC-703: Edit relationship dupe screen

james-baker-moj:
When editing the relationship type, if we receive a conflict from the API then show a new screen which shows the existing relationships between a prisoner and contact with an option to go to the duplicate record or contact list.


server/@types/contactsApi/index.d.ts
```diff
@@ -6824,7 +6824,7 @@ export interface components {
       isActive: boolean
     }
     PagedModelPrisonerContactSummary: {
-      content?: components['schemas']['PrisonerContactSummary'][]
+      content?: components['schemas']['c'][]
       page?: components['schemas']['PageMetadata']
```
timharrison-moj, line 6828: A `c` type? Unfinished typing, or genuinely a type called 'c'?

timharrison-moj (approved):
LGTM.

PR #458, opened 2025-05-12
PRC-735: Show inactive contacts by default

james-baker-moj:
We recently discovered that a lot of the contacts added by prisons are inactive. This is due to how adding contacts from the visitor screen in NOMIS works and may not be intentional. It may be that the active flag is not really used in day-to-day operations.  Show all contacts by default to allow more research to take place.


timharrison-moj (approved):
LGTM

PR #459, opened 2025-05-12
PRC-703: Handle duplicate on add contact

james-baker-moj:
If we get a 409 from the API then there was a duplicate contact. Give the user the option to go the existing relationship or to the contact list or back to check answers.


timharrison-moj (approved):
LGTM

PR #468, opened 2025-05-16
PRC-735: UI tweaks/fixes now we show inactive by default

james-baker-moj:
Clear filters was resetting to active only specifically rather than the default and some wording when there were no results was no longer accurate.


timharrison-moj (approved):
LGTM

PR #472, opened 2025-05-20
PRC-649: Show border under phase banner

james-baker-moj:
For some reason there was a custom style imported from the template project that has a custom header for the phase banner.


timharrison-moj (approved):
LGTM

PR #541, opened 2025-08-07
Hide dob for internal contact, and add delete-dob for non-internal contact

bill-moj:
PRC-877
for non-internal contact, show dob:
image

show change dob link only if there is no dob:
image

show change dob link and delete dob link otherwise:
image

PRC-878
do not show dob for internal contact:
https://files.slack.com/files-pri/T01066BNN9G-F098WU1GBJB/image.png

image


timharrison-moj (approved):
LGTM

PR #554, opened 2025-08-21
PRC-867 : remove internal check for caseload , now it's done in permission library

prabash-moj:


integration_tests/mockApis/componentApi.ts
```diff
@@ -20,6 +20,13 @@ const componentsNoServices: HeaderFooterSharedData = {
       type: 'INST',
       caseloadFunction: 'GENERAL',
     },
+    {
+      caseLoadId: 'KMI',
+      currentlyActive: false,
+      description: 'HMP Hewell',
+      type: 'INST',
```
timharrison-moj, line 27: `HMP Hewell` description appears in 2 x caseloads. Might be nicer to alter the description too, even if it doesn't affect the outcome of tests?

PR #568, opened 2025-09-15
PRC-894 : Create an alerts and restrictions page

prabash-moj:
PRC-894 : Create an alerts and restrictions page #568


scripts/generate-hmpps-contacts-api-types.sh
```diff
@@ -1,2 +1,2 @@
-npx -y openapi-typescript https://personal-relationships-api-dev.hmpps.service.justice.gov.uk/v3/api-docs | sed "s/\"/'/g" | sed "s/;//g" > ../server/@types/contactsApi/index.d.ts
+npx -y openapi-typescript http://localhost:8080/v3/api-docs | sed "s/\"/'/g" | sed "s/;//g" > ../server/@types/contactsApi/index.d.ts
 eslint --fix "../server/@types/contactsApi/index.d.ts"
```
timharrison-moj, line 2: Should probably still reference the DEV instance here?

server/views/pages/contacts/restrictions/reviewRestrictions.njk
```diff
@@ -0,0 +1,45 @@
+{% extends "../../../partials/layout.njk" %}
+{% from "govuk/components/button/macro.njk" import govukButton %}
+{% from "govuk/components/radios/macro.njk" import govukRadios %}
+{% from "partials/restrictionsReviewTable.njk" import restrictionsReviewTable %}
+
+{% set mainClasses = "app-container govuk-body" %}
+
+{% set pageTitle = "Review prisoner and contact restrictions - Manage contacts - DPS" %}
+{% block content %}
```
timharrison-moj, line 9: Seems like an odd page title?  This correct?

timharrison-moj (dismissed):
Couple of suggestions / comments to think on, but otherwise LGTM.

server/routes/contacts/manage/list/listPrisonerRestrictionsAlertsController.test.ts
```diff
@@ -0,0 +1,74 @@
...
+      alertsService.getAlerts.mockResolvedValue(alertsContent)
+
+      // When
+      const response = await request(app).get(`/prisoner/${prisonerNumber}/alerts-restrictions`).expect(200)
+
+      // Then
+      const $ = cheerio.load(response.text)
+      expect($('title').text()).toContain('Review prisoner and contact restrictions')
+      expect($('[data-qa="CONTACT_GLOBAL-title"]').text()).toContain('Global restrictions')
+      expect($('[data-qa="prisoner-alerts-heading"]').text()).toContain('John Smith’s prisoner alerts')
```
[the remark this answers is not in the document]
prabash-moj: Fair point, it is just because I was following the same pattern of the other view restrictions page, I think this can move to a unique id, as it is only used to identify the element in HTML for Cypress, Cheerios tests, etc.

scripts/generate-hmpps-contacts-api-types.sh
prabash-moj: My bad, will remove this, was testing UI changes before the api changes merged into dev.

server/views/pages/contacts/restrictions/reviewRestrictions.njk
timharrison-moj: It probably should have a title, was just questioning the length of it, rather than its removal. Probably best to reinstate so something is present, and we can check in dev what other page titles around it look like?

prabash-moj: my bad, forgot to update it here, I updated it in the code to the following text, 'Review prisoner restrictions and alerts - Manage contacts - DPS' 
image
And it is the title shown on the browner tab when you move the move over the tab, so this title here would be ok?

PR #580, opened 2025-09-23
PRC-894 : create an alerts and restrictions page   

prabash-moj:
Adding alerts and restrictions page,
Test evidence, 
image


timharrison-moj (approved):
LGTM

PR #592, opened 2025-09-30
Reapply "PRC-894 : create an alerts and restrictions page    (#580)" (#590)

prabash-moj:


server/routes/contacts/manage/relationship/comments/manageRelationshipCommentsController.test.ts
```diff
@@ -86,6 +89,15 @@ describe('GET /prisoner/:prisonerNumber/contacts/manage/:contactId/relationship/
   })
 
   it('GET should block access without edit contacts permission', async () => {
+    app = appWithAllRoutes({
+      services: {
+        auditService,
+        prisonerSearchService,
+        contactsService,
+        alertsService,
+      },
```
timharrison-moj, line 98: Due to the number of similar tests, I wonder if a single method to setup the appWithAllRoutes might be a good idea? Not for this PR, but something to consider so there's one place to change when services change?

PR #778, opened 2026-04-09
Fix incorrect escaping on notification success banner

hutcheonb-moj:
Banner being passed around before adding to title was causing escaping issue

Removed 'single apostrophe' cypress test - was largely uneccessary


tpmcgowan (approved):
LGTM

PR #781, opened 2026-04-16
VB-6610 - Updates restrictions notification to GOV.UK Tag

simonnebMOJ:
Minor improvement to align restriction notification to use same components.  Also removes word "active" as that wasnt accurate.

### DIFF
--- a/server/routes/contacts/add/contact-match/contactMatchController.test.ts
+++ b/server/routes/contacts/add/contact-match/contactMatchController.test.ts
@@ -121,7 +121,7 @@ describe('Contact details', () => {
       expect($('[data-qa=confim-title-value-top]').text().trim()).toStrictEqual(
         'Check and confirm if you want to link contact Jones Mason to prisoner John Smith',
       )
-      expect($('.govuk-tag').first.text().trim()).toStrictEqual('Restrictions in place')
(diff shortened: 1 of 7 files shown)


server/routes/contacts/add/contact-match/contactMatchController.test.ts
```diff
@@ -121,7 +121,6 @@ describe('Contact details', () => {
       expect($('[data-qa=confim-title-value-top]').text().trim()).toStrictEqual(
         'Check and confirm if you want to link contact Jones Mason to prisoner John Smith',
       )
-      expect($('.moj-badge').text().trim()).toStrictEqual('Active restrictions in place')
```
psoleckimoj, line 124: Rather than removing don't you want to replace the relavent line with whatever CSS selector you need for the tag and the new text it uses?

psoleckimoj: I suspect you can just replace `.moj-badge` with `.govuk-tag` and update the text.

simonnebMOJ: Should now be fixed

tpmcgowan (approved):
LGTM!

PR #792, opened 2026-04-29
VB-6518 Handle alerts or restrictions APIs being unavailable

tpmcgowan:
The prisoner 'mini-profile' used throughout the service displays counts of alerts and restrictions. If API calls to get either of these fails the code would default to (incorrectly) rendering zero counts for alerts/restrictions.

This change improves handling of API failures by showing an inset error message explaining that alerts and/or restrictions information is unavailable.

### DIFF
--- a/server/middleware/populatePrisonerDetailsIfInCaseload.ts
+++ b/server/middleware/populatePrisonerDetailsIfInCaseload.ts
@@ -1,10 +1,7 @@
-import { Request, Response, NextFunction } from 'express'
+import { NextFunction, Request, Response } from 'express'
 import { PrisonerSearchService, ContactsService } from '../services'
 import { Prisoner } from '../data/prisonerOffenderSearchTypes'
 import { PrisonerDetails } from '../@types/journeys'
-import { PagedModelPrisonerRestrictionDetails } from '../@types/contactsApiClient'
-import logger from '../../logger'
-import { PageAlert } from '../data/alertsApiTypes'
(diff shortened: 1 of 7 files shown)


server/middleware/populatePrisonerDetailsIfInCaseload.ts
```diff
@@ -17,43 +14,32 @@ const populatePrisonerDetailsIfInCaseload = (
     const { prisonerNumber } = req.params
     const { user } = res.locals
 
-    try {
```
tpmcgowan, line 20: I've removed all the try/catch blocks here because (unhandled) errors just get passed to the app-level error handler.

server/routes/prisonerMiniProfile.test.ts
```diff
@@ -0,0 +1,195 @@
...
+  prisonerSearchService.getByPrisonerNumber.mockResolvedValue(prisoner)
+})
+
+afterEach(() => {
+  jest.resetAllMocks()
+})
+
+// Mini profile displayed on most contacts pages; testing here on
+// GET /prisoner/:prisonerNumber/contacts/list
+describe('Prisoner Mini Profile', () => {
```
tpmcgowan, line 62: Although mini profile on every page there were no existing tests for it. So added full range of tests in addition to those covering alerts/restrictions fetch failure.

hutcheonb-moj (approved):
lgtm

PR #830, opened 2026-06-30
Fix copy for Date of Birth sort-by option

sorrelingabout:
Swapped "most recent" and "oldest" copy to correctly align with asc/desc sort-by-order for age. 

Also adjusted phrasing  to make it clearer.


tpmcgowan (approved):
LGTM

PR #831, opened 2026-07-03
Fix unescaped html fields

sorrelingabout:


server/utils/formatName.ts
```diff
@@ -18,7 +18,7 @@ const formatNameLastNameFirst = (
       name += ` ${val.middleNames}`
     }
   }
-  return capitaliseName(name)
+  return escapeHtml(capitaliseName(name))
```
tpmcgowan: I haven't tested but I wonder whether always escaping the name here means that in cases where it is correctly escaped in the template it will be potentially be double-escaped and look messy? E.g. [here](https://github.com/ministryofjustice/hmpps-contacts-ui/blob/64387217831de87f1c67d420d7535011f374292f/server/views/partials/navigation.njk#L18) in the breadcrumbs.

server/views/pages/contacts/manage/contactDetails/confirmDeleteIdentity.njk
```diff
@@ -20,7 +20,7 @@
                                 text: "Document type"
                             },
                             value: {
-                                html: identityDocument.identityTypeDescription,
+                                html: identityDocument.identityTypeDescription | escape,
```
tpmcgowan, line 23:
```suggestion
                                text: identityDocument.identityTypeDescription,
```

The GOV.UK components generally allow values to be passed to their macros as `text` or `html`. E.g. see docs for this one which is the [Summary list](https://design-system.service.gov.uk/components/summary-list/).

In our HMPPS repos we have Nunjuck's auto-escaping [turned on](https://github.com/ministryofjustice/hmpps-contacts-ui/blob/64387217831de87f1c67d420d7535011f374292f/server/utils/nunjucksSetup.ts#L80). This means no need to escape something rendered like `{{ contactName }}`.

In the GOV.UK components they [explicitly render](https://github.com/alphagov/govuk-frontend/blob/cfd2224301604372b0f66757c0ee0838d8c0a87c/packages/govuk-frontend/src/govuk/components/summary-list/template.njk#L65) `html` fields **without** escaping which is the issue. The `text` properties benefit from the auto-escaping.

I think fundamentally the problem in this repo is that values are passed to macros as `html` when just using `text` sorts the problem. (Only reason to use `html` would be if, say, needing to wrap a contact name in a span for styling or something...)

tpmcgowan (commented):
Potential side effect of double-escaping (untested) and may be a simpler fix

server/views/partials/linkedPrisonersTable.njk
```diff
@@ -9,7 +9,7 @@
   {% set rows = [] %}
   {% for linkedPrisoner in linkedPrisoners %}
     {% set prisonerNameAndNumberHtml %}
-      {{ linkedPrisoner | formatNameLastNameFirst }}<br />
+      {{ linkedPrisoner | formatNameLastNameFirst | escape }}<br />
```
tpmcgowan: Not tested but I don't think you need `| escape` here because rendering it as `{{ variable }}` will mean Nunjucks auto-escapes - so it should be safe to then feed into the `html` on line 17. I think adding the escape here might cause double-escaping.

server/views/pages/contacts/add/existing/match.njk
```diff
@@ -33,7 +33,7 @@
 
       {% if not isGranted(PersonalRelationshipsPermission.edit_contact_restrictions, prisonerPermissions) %}
        {{ govukWarningText({
-          html: 'As a Contacts Administrator, you cannot approve ' + (contact | formatNameFirstNameFirst) + ' to visit ' + ( prisonerDetails | formatNameFirstNameFirst(excludeMiddleNames = true) ) + '. If you need to do this, ask for the Contacts Authoriser role.',
+          html: 'As a Contacts Administrator, you cannot approve ' + (contact | formatNameFirstNameFirst | escape) + ' to visit ' + ( prisonerDetails | formatNameFirstNameFirst(excludeMiddleNames = true) ) + '. If you need to do this, ask for the Contacts Authoriser role.',
```
tpmcgowan: In this concatenation the first `formatNameFirstNameFirst` is escaped but not the second one. A few other instances of this to check.

server/views/pages/contacts/manage/employments/checkEmployer/organisationAddress/macro.njk
```diff
@@ -106,7 +106,7 @@
       text: "Comments on this address"
     },
       value: {
-      html: (params.address.comments or 'Not provided') | escape | nl2br
+      text: params.address.comments or 'Not provided'
```
tpmcgowan: This was OK as it was I think

tpmcgowan: Need to put the `| nl2br` back otherwise addresses will render all on one line. Basically I think this one was fine as it was because it set the macro's `html` property but before doing so escaped the content and added line breaks

server/views/partials/contactDetails/prisonerContactSummaryTable.njk
```diff
@@ -51,7 +51,7 @@
     {% set classes = 'light-grey govuk-!-padding-top-3 govuk-!-padding-bottom-3 max-200-width' if not contact.isRelationshipActive else 'govuk-!-padding-top-3 govuk-!-padding-bottom-3 max-200-width' %}
     {% set rows = (rows.push([
       { html: contactNameAndNumber, classes: classes + ' govuk-!-padding-left-1'},
-      { text: contact | formatAge, classes: classes },
+      { html: contact | formatAge, classes: classes },
```
tpmcgowan: I think this can stay as `text`. The function just [returns](https://github.com/ministryofjustice/hmpps-contacts-ui/blob/7727e9a470f09240c27e5bfc7690c1e4b3e10537/server/utils/utils.ts#L153) plain text. So, no need to escape and pass as `html`.

server/views/pages/contacts/manage/employments/checkEmployer/organisationAddress/macro.njk
sorrelingabout: Yep, reverted

tpmcgowan (approved):
LGTM

PR #842, opened 2026-07-30
Make Contacts UI Proxy Aware

KristianGregor:
Coded using the proxy-awareness copilot skill for updating typescript UIs to align with template, make use of hmpps-typescript-lib where available, and prepare for proxying egress (at this stage in dev only)

Passes through all 4 stages of the proxy awareness approach, but migration to the hmpps azure-telemetry module is being handled separately, and this PR makes the telemetry proxy-capable directly instead. follow up PR will check over merits of upgrading to the telemetry lib. 

- Migrate alertsApiClient, prisonerSearchApiClient, organisationsApiClient, contactsApiClient, and prisonApiClient from bespoke superagent/agentkeepalive restClient.ts to @ministryofjustice/hmpps-rest-client's RestClient + asSystem().
- Delete obsolete server/data/restClient.ts.
- Add server/data/authenticationClient.ts as a temporary Phase 1 adapter that preserves existing token acquisition logic (full hmpps-auth-clients migration is Phase 2).
- Make hmppsAuditClient's SQSClient proxy-aware via new server/data/helpers/proxySupport.ts (NodeHttpHandler + HttpsProxyAgent), following the pattern from hmpps-typescript-lib PR #211. Used the local fallback pattern since hmpps-typescript-lib's main branch does not yet have a published audit-client package.
- Pin https-proxy-agent to ^7.0.0 (latest 9.x is ESM-only and breaks Jest/ts-jest).
- Add NODE_USE_ENV_PROXY, APPLICATION_INSIGHTS_NO_STATSBEAT and the hmpps-envoy-https-proxy-env namespace secret to helm_deploy/values-dev.yaml only (preprod/prod rollout deferred).
- Fix ~6 production call sites and their tests for the SanitisedError shape change (.status -> .responseStatus) introduced by hmpps-rest-client's SanitisedError, including server/errorHandler.ts which now checks both .status and .responseStatus for compatibility with http-errors-based errors.


KristianGregor:
This work has been incorporated into a separate PR that has been merged into main

PR #843, opened 2026-07-30
Migrate telemetry from applicationinsights to hmpps-azure-telemetry

KristianGregor:
Full OpenTelemetry-based telemetry migration, replacing the raw applicationinsights SDK with @ministryofjustice/hmpps-azure-telemetry, matching the current hmpps-template-typescript pattern. Separated out into its own PR as this upgrade was seen to be a significant change on top of everything already on the proxy awareness ticket. 

- server/utils/azureAppInsights.ts: now calls initialiseTelemetry(...) as a side-effecting bootstrap module (filtering health/ping/info/ assets/favicon paths, enriching span names with the HTTP route), and exports telemetryUserAttributesMiddleware() which sets username/activeCaseLoadId as span attributes, replacing the old addUserDataToRequests App Insights telemetry processor, which had no direct equivalent since OTel span modifiers cannot see req/res.
- server.ts: imports azureAppInsights as the very first statement (before anything else, including the express-importing app module) so OTel instrumentation can patch http/express/bunyan before they are first required; adds SIGTERM/SIGINT handlers that flush telemetry before exiting.
- server/services/telemetryService.ts: drops the applicationInsightsClient constructor param and calls telemetry.trackEvent(...) directly, sanitising null/undefined property values (the new library's attribute type does not allow them).
- server/services/index.ts / server/data/index.ts: remove initialiseAppInsights()/buildAppInsightsClient() and the applicationInsightsClient plumbing; pass the new telemetry object straight to PrisonPermissionsService.create() as telemetryClient.
- server/app.ts: removes appInsightsMiddleware(), adds telemetryUserAttributesMiddleware() after setUpCurrentUser() (so it runs after the active caseload has been populated).
- package.json: removes applicationinsights (only kept previously for its TelemetryClient type via hmpps-prison-permissions-lib); upgrades @ministryofjustice/hmpps-prison-permissions-lib to 4.2.0, which drops applicationinsights as a dependency in favour of a minimal structural telemetryClient interface; adds @ministryofjustice/hmpps-azure-telemetry@1.0.2; adds @opentelemetry/api and @opentelemetry/resources overrides to resolve a peer-dependency clash with @sentry/node's bundled OpenTelemetry stack.

Validated: typecheck, lint, full unit test suite (188 suites / 2772 tests) all pass; npm ci succeeds cleanly against the updated lockfile; manually smoke-tested app boot, /health response, and graceful SIGTERM shutdown (confirmed telemetry flush runs before exit).


KristianGregor:
This work was incorporated into a different PR that has been merged into main

PR #851, opened 2026-08-07
Proxy awareness opentelemetry

KristianGregor:
Full upgrade to all available hmpps typescript libraries


package.json
```diff
@@ -114,7 +113,9 @@
     "typescript": "^6.0.3"
   },
   "overrides": {
+    "@opentelemetry/api": "1.9.0",
     "@opentelemetry/core": "2.10.0",
+    "@opentelemetry/resources": "2.10.0",
```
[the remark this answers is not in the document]
KristianGregor: dep graph resolves cleanly at present, this override may not be necessary but I think it can be addressed in subsequent changes if the dep tree becomes a problem

server/utils/azureAppInsights.ts
```diff
@@ -1,77 +1,33 @@
...
-}
+initialiseTelemetry({
+  serviceName: applicationName,
+  serviceVersion: buildNumber,
+  connectionString: process.env.APPLICATIONINSIGHTS_CONNECTION_STRING,
+  debug: process.env.DEBUG_TELEMETRY === 'true',
+})
+  .addFilter(telemetry.processors.filterSpanWherePath(['/health', '/ping', '/info', '/assets/*', '/favicon.ico']))
+  .addModifier(telemetry.processors.enrichSpanNameWithHttpRoute())
+  .startRecording()
```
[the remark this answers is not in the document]
KristianGregor: meh, we never run without that connection string set anyway

server/data/alertsApiClient.test.ts
```diff
@@ -14,7 +14,7 @@ describe('Alerts api client tests', () => {
 
   beforeEach(() => {
     fakeAlertsApi = nock(config.apis.alertsApi.url)
-    alertsApiClient = new AlertsApiClient()
+    alertsApiClient = new AlertsApiClient(new AuthenticationClient(config.apis.hmppsAuth, console))
```
psoleckimoj, line 17: Not sure why this needs to take `console`? It's not a valid arg in the `AlertsApiClient` constructor?

psoleckimoj: Ditto for all the other tests...

server/data/contactsApiClient.ts
```diff
@@ -64,7 +64,7 @@ export default class ContactsApiClient extends RestClient {
         path: `/contact`,
         data: request,
       },
-      user,
+      asSystem(user.username),
```
psoleckimoj: Since you're doing such a major refactor, I'd update these functions to just take `username` as that's all they need rather than the entire `user` object.  Same for all the other instances like this.

server/data/prisonApiClient.test.ts
```diff
@@ -1,29 +1,59 @@
...
+  const actual = jest.requireActual('node:fs')
+  return { ...actual, createReadStream: jest.fn(actual.createReadStream) }
+})
 
 const user = { token: 'userToken', username: 'user1' } as Express.User
 
 describe('Prison api client tests', () => {
   let prisonApiClient: PrisonApiClient
-  const prisonerThumbnail = jest.spyOn(RestClient.prototype, 'prisonerThumbnail')
+  const stream = jest.spyOn(RestClient.prototype, 'stream')
```
psoleckimoj: I changed this to `prisonerThumbnail` because nothing else uses `stream` and makes it a bit clearer what it is? But entirely personal view on this one.

server/data/prisonApiClient.ts
```diff
@@ -1,18 +1,44 @@
...
+      throw error
+    }
   }
 
   async getOffenderAddresses(prisonerNumber: string, user: Express.User): Promise<PrisonApiAddress[]> {
-    return this.get({ path: `/api/offenders/${prisonerNumber}/addresses` }, user)
+    return this.get({ path: `/api/offenders/${prisonerNumber}/addresses` }, asSystem(user.username))
+  }
+
+  // createReadStream() opens the underlying file descriptor asynchronously, so a missing/unreadable
```
psoleckimoj: Not sure we need such a massive comment, should be self-descriptive really. It's actually just mirroring the stream that the API returns whereas previously it couldn't do that and you just ended up with an error in the logs.

server/services/index.ts
```diff
@@ -19,16 +19,16 @@ export const services = () => {
...
     contactsApiClient,
     prisonApiClient,
     alertsApiClient,
     organisationsApiClient,
-    applicationInsightsClient,
   } = dataAccess()
 
   const auditService = new AuditService(hmppsAuditClient)
-  const telemetryService = new TelemetryService(applicationInsightsClient)
+  const telemetryService = new TelemetryService()
```
psoleckimoj, line 31: Just wondering if this wants replacing where `services()` is called from so fully replacing `applicationInsightsClient`?

server/data/alertsApiClient.test.ts
KristianGregor: It's not being passed to AlertsApiClient, its being passed to AuthenticationClient. hmpps-auth-clients AuthenticationClient accepts (Logger | Console) here. console is provided as a lightweight alternative to wiring in the bunyan logger, in a test context.

psoleckimoj: Cheers, misreading my commas 😆

server/data/contactsApiClient.ts
KristianGregor: done

server/data/prisonApiClient.ts
KristianGregor: deleted

